Issue #21, opened 2014-12-04
Use local version if available


MoOx:
Since -g should be avoided for node modules, I always install every tool I use on each project. So I've eslint for each project. Would be nice if that the linter-* can use local version if available (in /node_modules/eslint in this case).

That would allow for example to test new version like current @es6jsx branch :)

nelix:
Would it be possible to check the current projects package.json and then spawn the current projects ./node_modules/eslint in a new process? (I assume you can't just require it directly)
This way we could have customer plugins and be in line with the projects intended eslint version, it would also make updating eslint versions easier?

jfaissolle:
Yes, it is the only solution I see at the moment. See my comment in #26 which is related to this issue.

Issue #14, opened 2014-10-17
Support for .eslintignore


davidtheclark:
Currently it looks like the plugin ignores `.eslintignore` files. It would be nice if it would use the file instead.

(If I inadvertently open minified jQuery, for example, the linting crashes Atom --- I'd like to be able to not lint it inadvertently when I open it inadvertently.)

Issue #11, opened 2014-10-05
Use eslint CLIEngine API instead of command line


jfaissolle:
For the eslint team, the `CLIEngine` API should be the preferred way to integrate.

Issue #9, opened 2014-09-04
Match eslint CLI behavior with regard to .eslintrc files in parent dirs


focusaurus:
From what I can tell, linter-eslint seems to only find the nearest `.eslintrc` file, whereas the eslint CLI walks the filesystem upward and combines multiple files. For example, I do:

```
~/project/.eslintrc #project-wide config
~/project/test/.eslintrc #test-config, mocha env, etc
~/project/server/.eslintrc #node env, etc
```

With regular `eslint`, my pwd can be anywhere (including completely outside of ~/project) and all the .eslintrc files are found and properly interpreted.

If there's an eslint javascript API that directly matches just running `eslint path/to/some/file`, I'd rather have this plugin do that without adding any special rules or requirements.

PR #26, opened 2015-01-13
Make path to executable configurable

zachasme:
I wanted to try out the `es6jsx` branch on eslint, but I was unable to figure out how to patch in a configurable executable path without removing your custom `lintfile()`.

This pull request uses the base linter logic of Atom Linter instead of the previous custom `lintFile` method, and includes a new option for configuring the path to the `eslint` executable.

### DIFF
--- a/lib/init.coffee
+++ b/lib/init.coffee
@@ -2,12 +2,16 @@ path = require 'path'
 
 module.exports =
   config:
-    eslintRulesDir:
+    eslintExecutablePath:
       type: 'string'
-      default: ''
+      default: path.join __dirname, '..', 'node_modules', 'eslint', 'bin'
     disableWhenNoEslintrcFileInPath:
       type: 'boolean'
       default: false
+    eslintRulesDir:
+      description: 'Relative to working directory'
+      type: 'string'
+      default: ''
 
   activate: ->
     console.log 'activate linter-eslint'
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -1,87 +1,55 @@
-linterPath = atom.packages.getLoadedPackage('linter').path
+linterPath = atom.packages.getLoadedPackage("linter").path
 Linter = require "#{linterPath}/lib/linter"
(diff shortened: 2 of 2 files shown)


jfaissolle:
Hello @zachasme, thanks for your interest in the project. 
The original code of linter-eslint (before #11) was in fact similar to what you propose. It was based on the command line eslint tool and you could select the executable path in the config. This was changed for #9 because specifying the config in with the `--config` option does not match the cli tool behavior. By applying your PR, we would lose at least #9 and #14 fixes... So sorry, but I have to reject it...
Your problem is nonetheless a real one. It is also linked to #21, for which a solution would be really cool. I was thinking of a two steps approach. The linter-eslint plugin could launch another script (with exec) with a trick like https://www.npmjs.org/package/app-module-path to use the intended eslint API implementation.

dmnd:
@zachasme a really simple way to try out the new branch is to modify package.json:

```
dmnd@synapse:~/github/linter-eslint 🌱 master *= 
$ git diff
diff --git i/package.json w/package.json
index dd12aba..8821774 100644
--- i/package.json
+++ w/package.json
@@ -11,6 +11,6 @@
     "atom": ">0.50.0"
   },
   "dependencies": {
-    "eslint": "^0.10.1"
+    "eslint": "eslint/eslint#es6jsx"
   }
 }
```

zachasme:
Fair enough. Would a pull request on eslint adding a --configdir/--cwd/etc parameter for setting the path from which to search upwards be a better solution?

jfaissolle:
This would solve only a part of the problem. The #14 issue was about the use of .eslintignore files. If we use the command line tool, the --configdir or whatever option would not be sufficient to determine if the file should be linted or not. We would still need to use the eslint API for this... So I would prefer a solution fully using the API. Something like this could work: 
- launch a script more or less like this:
```
process.exec(['node', __dirname + '/eslintBootstrap.js', eslintDir])
```
- where  the script does 
```
require('app-module-path').addPath(process.argv[2]);
// lint and output to stdout
```

PR #29, opened 2015-02-08
Support predefined Meteor globals (update eslint dependency)

johanbrook:
Update to eslint `^0.14.0`, which adds predefined [Meteor](http://meteor.com) globals.


johanbrook:
Thanks!

PR #32, opened 2015-02-22
Use resolve to locate a local version of eslint

AsaAyers:
Fixes https://github.com/AtomLinter/linter-eslint/issues/21
Fixes https://github.com/AtomLinter/linter-eslint/issues/26

[resolve](https://www.npmjs.com/package/resolve)
> implements the node require.resolve() algorithm such that you can require.resolve() on behalf of a file asynchronously and synchronously

This branch will find and load the eslint that is local to the project you're working on if it's available. If none is found it just falls back to the one packaged in `linter-eslint`. I did notice that it doesn't find a globally installed eslint. I'm new to Atom, but I suspect it's because this this runs from inside atom instead of from node.

This is the same technique that [grunt-cli](https://github.com/gruntjs/grunt-cli/blob/master/bin/grunt#L33) uses to load your local version of grunt, and I adopted it for [CoffeeLint](https://github.com/clutchski/coffeelint/blob/master/bin/coffeelint#L29)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ Linter package must be installed in order to use this plugin. If Linter is not i
 $ apm install linter-eslint
 ```
 
+linter-eslint will look for a version of eslint local to your project and use it if it's available. If none is found it will fall back to the version it ships with.
+
+Lets say you depend on a specific version of eslint, maybe it has unreleased features, maybe it's just newer than what linter-eslint ships with. If `your-project/node_modules/eslint` exists linter-eslint will try to use that.
+
 ## Settings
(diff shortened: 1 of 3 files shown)


jfaissolle:
Hello, thanks a lot for this PR. It looks great ! I will just do a few test and merge it ASAP.

jfaissolle:
Great work ! Can I ask you to add a short text in the README to make the behavior clear to the users ?

AsaAyers:
I think my explanation feels a bit awkward, but I think it explains what's going on.

PR #42, opened 2015-03-16
Fix ignorePath on Windows

burnnat:
The `.eslintignore` settings do not work on Windows because the relative path construction assumes the path separator is always forward-slash while Windows uses a backslash.


jfaissolle:
Thanks !

PR #46, opened 2015-03-18
Get CLIEngine to load plugins

### LINKED ISSUE #45: linter-eslint doesn't seem to load plugins
I setup https://github.com/yannickcr/eslint-plugin-react and now I get errors in Atom that don't show up when I run the eslint from the command line. I started looking at the source, I suspect the problem is that eslint expects to be run from `node` but when run from inside `atom` `require()` is a little different.

I'm looking into how to fix this, but wanted to open an issue for it first.

AsaAyers:
Fixes #45

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -16,11 +16,14 @@ class LinterESLint extends Linter
   linterName: 'eslint'
 
   _requireEsLint: (filePath) ->
+    @localEslint = false
     try
       eslintPath = resolve('eslint', {
         basedir: path.dirname(filePath)
       })
-      return require(eslintPath)
+      eslint = require(eslintPath)
+      @localEslint = true
+      return eslint
     # Fall back to the version packaged in linster-eslint
     return require('eslint')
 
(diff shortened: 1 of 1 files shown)


RickEyre:
+1

jfaissolle:
@AsaAyers, plugins were disabled because they couldn't be loaded with the version of `eslint` packaged with `linter-eslint`. With your fix, plugins can now be loaded but only with the local eslint, at least this is my understanding. You could keep the disable behavior when the plugin loads the packaged `eslint`. If you don't, some users will compain that there is a regression because they will suddenly have errors with their setup.

AsaAyers:
You were correct, if you are using the `eslint` packaged in `linter-eslint` it won't be able to find the plugins. I have restored the old behavior with the addition that it adds a warning on line 1 providing basic instructions on how to make it load your plugin.

Here is the warning I get when I configure the `react` plugin but `linter-eslint` has to disable it.

    `npm install eslint` in your project to enable plugins: react (linter-eslint)

lib/linter-eslint.coffee
```diff
@@ -56,18 +59,34 @@ class LinterESLint extends Linter
...
+          delete config.rules[key] if isPluginRule.test(key)
 
     result = linter.verify @editor.getText(), config
 
+    if config.plugins?.length and not @localEslint
+      result.push({
+        line: 1
+        col: 0
+        error: false
+        warn: true
```
jfaissolle: Should be `warning: true` to appear as warning instead of error.

jfaissolle:
Please amend the last commit to make the message appear as a warning.

AsaAyers:
It's fixed now. I was just looking for it to show up when I tested. Didn't notice it wasn't a warning.

jfaissolle:
Sorry, I read the code too fast... My solution was wrong. In order to make it appear as a warning, you have to do the following instead:

```coffeescript
result.push {
  line: 1
  col: 0
  severity: 1
  message: "`npm install eslint` in your project to enable plugins:
  #{config.plugins.join(', ')} (linter-eslint)"
}
```

AsaAyers:
Good catch. I also should have used `column` instead of `col`. I'm not sure if the column actually makes a difference for this.  ¯\\_(ツ)_/¯

This time I verified that it does look like a warning now.

PR #53, opened 2015-04-02
allows for linting of *.jsx files instead of *.js.jsx

aamaliaa:
Fixes support for *.jsx files as ESLint now supports it.


iam4x:
This is incorrect, you need `atom-react` plugin to have the syntax of JSX recognized.

See: https://github.com/orktes/atom-react/blob/master/settings/JavaScript%20(JSX).cson

The syntax is correctly called `.source.js.jsx`

aamaliaa:
Ah, thanks for the heads up. I was using the `language-jsx` package for some reason and it wasn't working but now that I've switched it works.

PR #60, opened 2015-04-10
Use activationCommands instead of activationEvents.

rosston:
Kills a Deprecation Cop warning.

activationEvents has been deprecated since v0.153.0:
https://github.com/atom/atom/commit/53b538311e85b2539e8709ef8dd44d8ab74e
9bab


iam4x:
Thank's :+1:

PR #69, opened 2015-04-29
Support `source.babel` syntax

gabrielecirulli:
This adds support for the `source.babel` syntax, provided by the package [language-babel](https://atom.io/packages/language-babel).


iam4x:
Thank's @gabrielecirulli

PR #76, opened 2015-05-13
Allow using global eslint and plugins via togglable config option

yamalight:
@iam4x here ya go - new mergeable version with config option :)


iam4x:
Add the option in `lib/init.coffee` and it will be perfect!

yamalight:
@iam4x done!

PR #77, opened 2015-05-15
Fix loading plugins while using globally installed eslint

yamalight:
@iam4x a small fix for plugins loading and global eslint


iam4x:
Thank's @yamalight :)

yamalight:
sweet! :+1:

PR #122, opened 2015-06-24
Adds semantic javascript to supported grammars

calebmer:
Many people use the [language-javascript-semantic](https://github.com/p-e-w/language-javascript-semantic) package in combination with this linter plugin. If we could have baked in support for the grammar in this plugin it would save the hassle of having to re-add it whenever the plugin updates.


iam4x:
Didn't know about this. Thank's for PR.

PR #124, opened 2015-06-25
Fix loading plugins while using global npm path

yamalight:
Current version does not loads plugins from global packages properly. This PR fixes this problem.


iam4x:
NIce easy fix, thank's @yamalight  :+1:

yamalight:
sweet, thanks!

PR #126, opened 2015-06-25
Add config option to show eslint ruleid in messages

Fjandin:
I have added a config option 'showRuleIdInMessage', that when set to
true will add the corresponding eslint rule id to warning/error
messages. Config option is default false.


iam4x:
Thank's :+1:

PR #139, opened 2015-07-06
Fix "fs is not defined" on linting

asan:
Hi, you use `fs` package on [L75](https://github.com/AtomLinter/linter-eslint/blob/master/lib/linter-eslint.coffee#L75) and it not properly requered, so i get "fs is not defined" error.

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -1,7 +1,7 @@
 path = require 'path'
 {sync} = require 'resolve'
 {execSync} = require 'child_process'
-{statSync, existsSync} = require 'fs'
+{statSync} = require 'fs'
 {CompositeDisposable} = require 'atom'
 {allowUnsafeNewFunction} = require 'loophole'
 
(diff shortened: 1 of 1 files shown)


iam4x:
`existsSync` is deprecated from node API, can you use `statSync` ?

I dislike coffee because it doesn't break when errors aren't defined with linter :(

asan:
Fix it, so funny that eslint can catch this type of errors easily, but this plugin is coffee not es6. dogfooding FTW

lib/linter-eslint.coffee
```diff
@@ -72,7 +72,7 @@ module.exports =
         # Add showRuleId option
         showRuleId = atom.config.get 'linter-eslint.showRuleIdInMessage'
 
-        if rulesDir and fs.existsSync rulesDir
+        if rulesDir and statSync(rulesDir).isDirectory()
```
iam4x: You have to do:

```
try
  options.rulePaths = [rulesDir] if rulesDir and statSync(rulesDir).isDirectory()
```

Because `statSync(rulesDir)` will throw an error if it doesn't exists...

iam4x:
@asan Yeah, I might rewrite the code in ES6/7 to use `await / async` and use async everywhere.

peol:
Would be great to have this merged and a new version released if possible, currently had to disable in-editor linting :/

asan:
@iam4x now ok?

iam4x:
@asan Perfect! :+1:

PR #142, opened 2015-07-07
Remove usage of `fs` when checking for rulesDir

peol:
Current version crashes when the `rulesDir` setting is used.


peol:
Sorry, just saw the other PR.

PR #145, opened 2015-07-10
Search for eslint in parent directories and fix `this.binary` caching

ide:
This tells linter-eslint to recursively search parent directories for the eslint binary. This supports projects that share a eslint configuration between multiple subprojects:

```
.
├── Client
│   ├── client.js
│   └── package.json
├── Server
│   ├── server.js
│   └── package.json
└── package.json  # eslint, babel-eslint, eslint-plugins here
```

With this commit, if you open up the Client or Server directory in Atom, linter-eslint will find eslint in the parent directory's node_modules.

There was also a bug where setting `this.binary` would add a property to the global object instead of the exports, since `findESLint` was using unbound functions. Switched to arrow functions to fix this.

### DIFF
--- a/lib/linter-eslint.js
+++ b/lib/linter-eslint.js
@@ -50,14 +50,18 @@ export default {
   },
 
   findESLint(cwd) {
-    return new Promise(function (resolve) {
+    return new Promise(resolve => {
       // Return cached copy of binary path
       if (this.binary) return resolve(this.binary)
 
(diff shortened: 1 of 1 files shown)


ide:
Could you re-apply this change? The problem is that if you open the Client directory in my example above, `atom.projects` contains the path to the Client directory but not any of its parents.

PR #150, opened 2015-07-10
[2.3.0 on Windows] .eslintignore not being honored

clintwood:
This PR ensures the correct platform path separators are used.

Using incorrect separators causes at least [these lines](https://github.com/AtomLinter/linter-eslint/blob/master/lib/linter-eslint.js#L112-L113) to not replace the `cwd` from the file path and therefore on Windows `linter-eslint` always passes the full path to ESLint causing `.eslintignore` logic to fail.


iam4x:
@clintwood I always miss platform specific thing, thank's :+1:

clintwood:
:)  FWIW I'm really not a fan of Windows backslash - it bites me everyday!

PR #152, opened 2015-07-10
add 'Show File Ignored Warning' option

clintwood:
@iam4x, I've added an option to hide the annoying **_File ignored because of your .eslintignore file. Use --no-ignore to override._**  warning message for files ignored in the `.eslintignore` file.

Since I didn't raise an issue yet please feel free to close this if this option is not what you want in `linter-eslint` :)

### DIFF
--- a/lib/linter-eslint.js
+++ b/lib/linter-eslint.js
@@ -21,6 +21,11 @@ export default {
       default: true,
       description: 'Show the `ESlint` rule before the issue message'
     },
+    showFileIgnoredWarning: {
+      type: 'boolean',
+      default: false,
+      description: 'Show "File ignored because of your .eslintignore file. Use --no-ignore to override." warning'
+    },
     disableWhenNoEslintrcFileInPath: {
       type: 'boolean',
       default: false,
@@ -138,6 +143,11 @@ export default {
       try {
         results = JSON.parse(output).results
         messages = results[0].messages || []
+        // clobber ignored message based on settings
+        if (messages.length === 1 && !!~messages[0].message.indexOf('ignored because of your .eslintignore file') &&
+            !atom.config.get('linter-eslint.showFileIgnoredWarning')) {
+          messages = []
+        }
       } catch (error) {
         console.warn('[Linter-ESLint] error while parsing ouput:')
         console.warn(output)


vramana:
@clintwood The proper fix is not to lint the files rather than adding an option to supress the message.

clintwood:
Totally agree but but now that you spawning a child process to eslint I'm not sure you can decide up front whether you should lint the file or not... unless you have access to the `.eslintignore` logic in `atom-linter`.

Also, other down sides to spawning out to a child process is that it's slower and I'm not sure you have access to the rule id...

clintwood:
@iam4x, I [see pressure](/AtomLinter/linter-eslint/issues/160) to get `linter-eslint` back to using the JS interface rather than spawning child process which I believe will enable the package to selectively lint based on the `.eslintignore` settings.

lib/linter-eslint.js
```diff
@@ -138,6 +143,11 @@ export default {
       try {
         results = JSON.parse(output).results
         messages = results[0].messages || []
+        // clobber ignored message based on settings
+        if (messages.length === 1 && !!~messages[0].message.indexOf('ignored because of your .eslintignore file') &&
```
iam4x: I really don't like `!!~` it doesn't make enough sense for everyone.

clintwood: Sure, We could match on the whole string... Would you prefer that? E.g. :
```js
 messages[0].message === 'File ignored because of your .eslintignore file. Use --no-ignore to override.'
```

clintwood:
I can update PR if you wish...

lib/linter-eslint.js
iam4x: I like this a lot: https://github.com/MoOx/eslint-loader/blob/master/index.js#L30 what do you think?

clintwood: Yes that's cleaner, although the whole idea of linting unnecessarily should be avoided.
I'll update PR...

clintwood:
To test PR I pulled latest upstream and I'm hitting [issue #166](https://github.com/AtomLinter/linter-eslint/issues/166) so linting is broken my side.  I think that would need to be fixed before updating this PR.

FWIW, I think the JS interface approach of using `eslint.CLIEngine(...)` [like done here](https://github.com/MoOx/eslint-loader/blob/master/index.js#L15) seems like the way to go...

clintwood:
@iam4x, I'm going to close this PR for now since the changes above are trivial to implement directly.  Hopefully this won't be needed if the JS API interface is used.

Issue #158, opened 2015-07-12
Allow setting of file extension (.jsx)


ludovico:
I am using eslint for both .js and .jsx (react) files. I may be wrong, but I can't see anywhere to add
extra file extensions to the settings. Am I missing something? I would be happy to implement it myself if it as actually missing, but I thought it better to ask here first.

ludovico:
I just added a pull request for it, simple as it is.

iam4x:
`.jsx` should be linted too, I'm using `source.js.jsx` to recognize the file syntax and process them into `eslint`

Do you have `react` or `babel` package installed into Atom?

ludovico:
I tried to go back to the official version and try it now, but I am having some trouble with it (unrelated). I am sure you are right and that the problem is on my end, even though I did have `react` installed. But I still think that it would be nice to be able to specify file extensions without any extra packages, if it doesn't slow the linting down too much.

iam4x:
I've rollbacked linter to 1.x with the 3.0.0 version published. Please open a new issue if it's still happening.

Make sure to quit Atom and `$ apm uninstall linter-eslint && apm install linter-eslint`.

PR #159, opened 2015-07-12
Add file extensions

ludovico:
I just opened an issue (#158), wanting to allow for extra file extensions. Seeing as it was very quickly done, I went ahead and added it. I don't really check the users input, but it shouldn't really be necessary. Hope it is acceptable.


iam4x:
This is not related to the actual codebase of the package, and like I said on the other issue `.jsx` are already linted because syntax is recognized from buffer.

PR #170, opened 2015-07-17
Search up the directory hierarchy for the local copy of eslint

ide:
This lets users open up subdirectories of their project and still use their project's copy of eslint. For example:

```
project/
  subdir/
    src.js
  node_modules/
    eslint/
```

If you open `subdir` in Atom, linter-eslint will now load `project/node_modules/eslint`.


iam4x:
Thank's @ide

PR #184, opened 2015-08-02
Cleanup

steelbrain:
Explaination
 - Doing linting in main thread introduced a significant lag for me on my not-so-old machine
 - Searching for plugins and configs and ignores in this package isn't really ideal as it does a lot of FS stats in the main thread and there's some `sync` stuff in the linter codebase
 - This makes this linter's codebase cleaner
 - This makes this linter future proof
 - Some redundant dependencies have been removed
 - We are still writing to stdin, not a file so there shouldn't be any performance downside

Down Sides:
 - ESLint's CLI api doesn't support reporting which rule generated that error, maybe we could PR to that repo?
 - Disable when no `.eslintrc` in file path is no longer available in favor of https://github.com/atom-community/linter/issues/744 which I'll be implementing **really** soon.

/cc @k2b6s9j @iam4x

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/184)


steelbrain:
I have better plans for this package. I'll be doing a child process, exchanging cli arguments, then requiring those modules there.

PR #185, opened 2015-08-03
Add option to select file extensions

sgtpepper43:
I haven't changed the readme yet because I want to get feedback on the best way to name/describe the options. My specific problem is I want to lint `.js.erb` files, but the `grammarScope` I need to add is `source.js.rails source.js.jquery` (don't ask me why). So that's why I've included both options, but I don't know how confusing that might be, so any feedback is appreciated.

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -31,6 +31,15 @@ module.exports =
       type: 'string'
       default: ''
       description: 'Run `$ npm config get prefix` to find it'
+    fileScopes:
+      type: 'string'
+      default: 'source.js,source.js.jsx,source.babel,source.js-semantic',
+      description: 'Lint files with these scopes (passed to atom-linter) (separated by commas)'
+    fileExtensions:
+      type: 'string'
+      default: '.js,.js.jsx,.babel,.js-semantic',
+      description: 'Lint files with these file extensions (separated by commas)'
+
 
   activate: ->
     console.log 'activate linter-eslint'
@@ -44,7 +53,7 @@ module.exports =
 
   provideLinter: ->
     provider =
-      grammarScopes: ['source.js', 'source.js.jsx', 'source.babel', 'source.js-semantic']
+      grammarScopes: atom.config.get('linter-eslint.fileScopes').split(',')
       scope: 'file'
       lintOnFly: true
       lint: (TextEditor) =>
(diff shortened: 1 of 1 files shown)


lib/linter-eslint.coffee
```diff
@@ -31,6 +31,15 @@ module.exports =
       type: 'string'
       default: ''
       description: 'Run `$ npm config get prefix` to find it'
+    fileScopes:
+      type: 'string'
```
Arcanemagus: This should be an [array](https://atom.io/docs/api/v1.1.0/Config#array) or strings.

lib/linter-eslint.coffee
```diff
@@ -31,6 +31,15 @@ module.exports =
...
+      default: 'source.js,source.js.jsx,source.babel,source.js-semantic',
+      description: 'Lint files with these scopes (passed to atom-linter) (separated by commas)'
+    fileExtensions:
+      type: 'string'
```
Arcanemagus: Same here.

lib/linter-eslint.coffee
```diff
@@ -44,7 +53,7 @@ module.exports =
 
   provideLinter: ->
     provider =
-      grammarScopes: ['source.js', 'source.js.jsx', 'source.babel', 'source.js-semantic']
+      grammarScopes: atom.config.get('linter-eslint.fileScopes').split(',')
```
Arcanemagus: You should be observing the setting in activate() instead of get()ing it every time.

lib/linter-eslint.coffee
```diff
@@ -78,6 +87,9 @@ module.exports =
             console.warn '[Linter-ESLint] ESlint rules directory does not exist in your fs'
             console.warn error.message
 
+        # Add file extensions option
+        options.extnensions = atom.config.get('linter-eslint.fileExtensions').split(',')
```
Arcanemagus: Is this even necessary? We are running executeOnText() instead of running on a file.

Arcanemagus:
@sgtpepper43 rebase on the current `master` and address the comments above and I suppose I don't see any major reason not to merge this.

lib/linter-eslint.coffee
steelbrain: Even if he gets everytime, the main function is invoked only once. User will have to deactivate/activate package for it to take effect. He needs to store `grammarScopes` pointer somewhere and then pass it to linter

Arcanemagus:
@sgtpepper43 since the comments haven't been addressed and a complete re-write of the linter has been pushed out since this was submitted I'm marking this as closed. Feel free to re-open for review if/when this gets redone for the current state of `linter-eslint`.

PR #188, opened 2015-08-06
Add small check to avoid undefined ruleId

btd:


Arcanemagus:
@btd can you provide an example that triggers this?

btd:
It already does not make sense, i never saw this again.

Issue #194, opened 2015-08-18
Lint within <script> tags in HTML files


klebba:
Plugins like this: https://github.com/BenoitZugmeyer/eslint-plugin-html won't work because linter-eslint ignores html files.

linter-jshint recently added this feature by modifying grammarScopes: https://github.com/AtomLinter/linter-jshint/issues/106

pitaj:
:+1:

pitaj:
ref #104

iam4x:
@klebba This is an easy PR could you do it to support HTML?

klebba:
@iam4x I wish I could but I'm on a deadline; it will be a while until I could actually get to it. Sorry I can't be of more help right now!

mehwww:
It seems that [linter](http://eslint.org/docs/developer-guide/nodejs-api.html#linter) wont work properly with plugins, so I create a PR use [executeOnText](http://eslint.org/docs/developer-guide/nodejs-api.html#executeontext) instead and extend grammarScopes make it work with HTML files

klebba:
@mehwww thanks!

PR #197, opened 2015-08-24
Lint javascript inside HTML file

mehwww:
#194 Use [executeOnText](http://eslint.org/docs/developer-guide/nodejs-api.html#executeontext) instead of [linter](http://eslint.org/docs/developer-guide/nodejs-api.html#linter) because `linter` does not support [plugin](https://github.com/BenoitZugmeyer/eslint-plugin-html)

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -27,6 +27,10 @@ module.exports =
       type: 'string'
       default: ''
       description: 'Run `$ npm config get prefix` to find it'
+    lintHtmlFiles:
+      type: 'boolean'
+      default: false
+      description: 'Enable lint JavaScript in HTML files'
 
   activate: ->
     require('atom-package-deps').install('linter-eslint')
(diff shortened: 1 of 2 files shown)


mehwww:
Maybe we should add a option for this feature?

yuche:
So helpful.Thanks.
@mehwww

ikeagold:
:+1: @iam4x pls, because its a last reason why many people still using jshint+jscs

Arcanemagus:
Have you done any testing on this? I would imagine that `eslint` is expecting you to feed it some JavaScript to lint, but this PR is simply feeding it the entire contents of the file you are editing. How would it handle mixed HTML and JavaScript, much less more complex combinations?

mehwww:
I personally use eslint to lint html files with [a eslint plugin](https://github.com/BenoitZugmeyer/eslint-plugin-html), so if you are planning to support this feature maybe you shall mention it in the document or setting things, or I guess there will be a lot of code do the spliting things

ikeagold:
Is eslint on ST3 won't lint inline js too? or its only Atom have this issue's? and same issue in jshint(maybe linter-xo too) from @sindresorhus .
@mehwww thanx for eslint-html, but can't find it on (atom.io/packages](https://atom.io/packages/)

ikeagold:
@mehwww maybe you rebase pull and make it on logic like [here](https://github.com/AtomLinter/linter-jshint/blob/master/lib/main.coffee#L25-L29) (from atom-jshint), this logic clean and looking ok) and working fine ;)

mehwww:
@ikeagold It's an eslint plugin but not an atom package, you need install it via `npm install eslint-plugin-html` locally or globally, see [here](http://eslint.org/docs/user-guide/configuring#configuring-plugins)
And it's fair enough making it more logic, but i'd rather let the maintainer decide how to support this feature, by default or provide a configure option or somewhat else

Arcanemagus:
@ikeagold The part of `linter-jshint` you linked has nothing to do with this beyond saying it can handle those scopes, the relevant part of that file is [here](https://github.com/AtomLinter/linter-jshint/blob/master/lib/main.coffee#L55) where it tells `jshint` to do the extraction for it.

I don't think requiring an `eslint` plugin is the way to go here since one of the large parts of this is that you can use a custom installation with no additional dependencies. If this is going to be supported the extraction would need to be done locally in this package in some manner.

Arcanemagus:
@mehwww It would be great to see a PR (maybe this one? :stuck_out_tongue:) that properly supports that, but it's more involved than what you have here.

Unless of course I'm totally wrong, I am only basing this on the description of `executeOnText()` and haven't verified it doesn't already work myself after all, hence asking if you had tested this :wink:.

mehwww:
@Arcanemagus I‘ve updated my PR, update the deps, drop support for old eslint and requrie the plugin locally and conditionally load it. It works well here and you may test this :wink:.

ikeagold:
@mehwww thanks man, brilliant job :+1:

iam4x:
@mehwww Going to test this :+1: 

Sorry for not being much present here. We raised 15M where I work, and now things are going at the office...

iam4x:
:ship:

Issue #202, opened 2015-08-30
Indent definition rule not found.


nwhite89:
So I'm looking at moving to `ESLint` from `JSHint` & `JSCS`, I'm getting the following error whilst using the `eslint-linter` for Atom and I'm getting an error stating that `indent` isn't an option?

Any help on this would be greatly appreciated.

Error: 
```
Definition for rule 'indent' was not found.
/Users/nickwhite/.atom/packages/linter-eslint/lib/linter-eslint.coffee:150 Error: Definition for rule 'indent' was not found.
  at /Users/nickwhite/Sites/static_html/node_modules/eslint/lib/eslint.js:591:27
  at Array.forEach (native)
  at EventEmitter.module.exports.api.verify (/Users/nickwhite/Sites/static_html/node_modules/eslint/lib/eslint.js:567:16)
  at /Users/nickwhite/Dropbox/Atom/packages/linter-eslint/lib/linter-eslint.coffee:111:18
  at exports.allowUnsafeNewFunction (/Users/nickwhite/Dropbox/Atom/packages/linter-eslint/node_modules/loophole/lib/loophole.js:25:14)
  at Object.provider.lint (/Users/nickwhite/Dropbox/Atom/packages/linter-eslint/lib/linter-eslint.coffee:109:13)
  at Promise.then._this.emitter.emit.linter (/Users/nickwhite/Dropbox/Atom/packages/linter/lib/linter-registry.coffee:57:22)
  at LinterRegistry.triggerLinter (/Users/nickwhite/Dropbox/Atom/packages/linter/lib/linter-registry.coffee:56:16)
  at /Users/nickwhite/Dropbox/Atom/packages/linter/lib/linter-registry.coffee:50:17
  at Array.map (native)
  at /Users/nickwhite/Dropbox/Atom/packages/linter/lib/linter-registry.coffee:48:27`
```

My config file looks like:
```
{
  "env": {
    "browser": true
  },
  "globals": {},
  "rules": {
    "no-debugger": 0,
    "no-eval": 0,
    "dot-notation": 0,
    "eqeqeq": 0,
    "indent": [2, 2, {
        "VariableDeclarator": 2,
        "SwitchCase": 2
    }],
    "camelcase": [
      2,
      {
        "properties": "never"
      }
    ],
    "curly": [
      2,
      "all"
    ],
    "space-return-throw-case": 0,
    "space-after-keywords": [
      2,
      "always"
    ],
    "space-before-blocks": [
[...]

PR #211, opened 2015-09-14
Support local gulp-eslint installations.

onlywei:
Addresses #202.

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -214,7 +214,12 @@ module.exports =
       try
         eslintPath = sync 'eslint', {basedir: currentPath}
       catch
-        continue
+        try
+          # Support gulp-eslint installations
+          gulpEslintPath = sync 'gulp-eslint', {basedir: currentPath}
+          eslintPath = sync 'eslint', {basedir: path.dirname gulpEslintPath}
+        catch
+          continue
       return allowUnsafeNewFunction -> require eslintPath
     throw new Error "Could not find `eslint` locally installed in #{ path.dirname filePath } or any parent directories"
 


IanVS:
Shouldn't the `eslint` within `gulp-eslint` already be found by `resolve.sync` as it is?

onlywei:
@IanVS It does not seem to work that way.

onlywei:
@IanVS For example, I cannot do `require('eslint')` directly in my local files, I can only do `require('gulp-eslint')`.

IanVS:
Guess so, at least until everyone is using `npm 3.0.0`, wherein:

> Dependencies will now be installed maximally flat. Insofar as is possible, all of your dependencies, and their dependencies, and THEIR dependencies will be installed in your project's node_modules folder with no nesting. You'll only see modules nested underneath one another when two (or more) modules have conflicting dependencies.

onlywei:
@IanVS I don't think we should wait for that to happen. AFAIK NPM 3.0.0 isn't even out yet.

IanVS:
It's in [beta](https://github.com/npm/npm/releases/tag/v3.3.3) right now, but I wasn't suggesting to wait for it.  It still seems like there should be a better way to accomplish this, but I don't know what it would be.

onlywei:
@iam4x do you have an opinion on this?

steelbrain:
Fixed in another way by #258

onlywei:
Installed v5.0.0. Tested. Works! Thanks!

MatthewHerbst:
This doesn't seem to be working properly anymore.

```
Atom v1.7.3
Atom linter-eslint v7.2.2
Node v4.2.6
npm v2.14.12
```
Locally installed/package.json
```
gulp-eslint v2.0.0
eslint-plugin-react v5.0.1
```

When I open a file in Atom, I get an error that the `eslint-plugin-react` module can't be found. This error is resolved by locally installing `eslint`.

Or did I miss something?

Arcanemagus:
@MatthewHerbst The discussion above is talking about taking advantage of a fluke in how `npm@^3` installs dependencies... if you want to have a guaranteed `eslint` available you should explicitly list it.

Supporting grabbing the sub-dependency of `gulp-eslint` in all possible situations isn't something that I feel should be supported in this package.

MatthewHerbst:
@Arcanemagus ah alright. That's fine by me, I just thought the above meant finding sub-dependency grabbing should be supported.

It could be worth making a UI option where a user could specify a path to a specific install of `eslint`

PR #217, opened 2015-09-25
Auto install linter

### LINKED ISSUE #207: Install `linter` automatically.
By using [`atom-package-deps`](https://github.com/steelbrain/package-deps) you are able to install `linter` automatically on activation. This will allow new Atom users to install only your package and nothing else, removing one step from the installation process.

Please see AtomLinter/linter-ruby#38 for example.

### LINKED ISSUE #212: Please do not require the `linter` package
Nuclide has released a [suite of packages](https://github.com/facebook/nuclide/tree/master/pkg/nuclide/diagnostics) that displays diagnostic results.

However, Nuclide users still [cannot use this package](https://github.com/facebook/nuclide/issues/190) because it [will not start](https://github.com/AtomLinter/linter-eslint/blob/master/lib/linter-eslint.coffee#L11) unless the `linter` package is installed.

Please remove that requirement, or at least relax it to allow the `nuclide-diagnostics` suite of packages to stand in for the `linter`.

Arcanemagus:
Automatically install the `linter` package, if it is not already. Remove the requirement that it be loaded to allow the use of an alternative consumer of this provider.

Closes #207.
Closes #212.

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/217)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,25 @@
 # linter-eslint
 
-This linter plugin for [Linter](https://github.com/AtomLinter/Linter) provides an interface to [eslint](http://eslint.org). It will be used with files that have the “JavaScript” syntax.
+This linter plugin for [Linter](https://github.com/AtomLinter/Linter) provides
+an interface to [eslint](http://eslint.org). It will be used with files that
+have the “JavaScript” syntax.
 
 ## Installation
-Linter package must be installed in order to use this plugin. If Linter is not installed, please follow the instructions [here](https://github.com/AtomLinter/Linter).
+```ShellSession
+apm install linter-eslint
+```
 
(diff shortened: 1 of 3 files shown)


Arcanemagus:
@nmote, @admmasters, @bolinfest

This should be a satisfactory compromise, it will install the `linter` package if it is not already installed, but will no longer force it to be enabled. Please test this and make sure that it does what you are looking for and solves the issue outlined in #212.

nmote:
@Arcanemagus: This looks like a good compromise to me. Thanks for putting this up.

steelbrain:
okay then.

PR #224, opened 2015-10-06
Capitalize types of linter messages

hansonw:
"error" and "warning" should be "Error" and "Warning" as per the official linter spec. (see https://github.com/atom-community/linter/wiki/Linter-API, in particular https://github.com/atom-community/linter/wiki/Registering-Custom-Message-Types)

This doesn't normally matter, but other consumers of the API (at the moment, namely Nuclide's diagnostics plugin) might be sensitive to the distinction.

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -116,14 +116,14 @@ module.exports =
 
                   if showRuleId
                     {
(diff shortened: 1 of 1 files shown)


iam4x:
:+1:

Issue #239, opened 2015-10-25
Add text field to error messages


benjaminRomano:
I currently work on the package linter-manager that   adds additional functionality to the message display. 

For my package, I'm not able to use the html property on the messages. Would it be possible to have the messages include the text property as well?

iam4x:
Closed with #240

PR #240, opened 2015-10-25
Added text property to errors

benjaminRomano:
Fix for #239.

The linter package will still display the html , but other packages that consume linter messages can choose between using either html or text as shown below.

image

For the text I made it follow this convention: ruleId - message. Does this work or would you prefer it be formatted in a different way?

### DIFF
--- a/lib/linter-eslint.coffee
+++ b/lib/linter-eslint.coffee
@@ -118,6 +118,7 @@ module.exports =
                     {
                       type: if severity is 1 then 'Warning' else 'Error'
                       html: '<span class="badge badge-flexible">' + ruleId + '</span> ' + message
+                      text: "#{ruleId} - #{message}"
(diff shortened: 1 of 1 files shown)


iam4x:
:ship: 

Thank's.

steelbrain:
IMO this is a violation of the linter API, because it either wants you to provide a text or an html. If you use both, then I can not guarantee a consistant behavior between versions, because in the internal codebase, we could at some places do something like
```js
if (message.text) {
  use message.text
} else if (message.html) {
  use message.html
}
```
and in other places it could be
```js
if (message.html) {
  use message.html
} else if (message.text) {
  use message.text
}
```

The consumers of the API should do this too instead of doing patches like this.
@iam4x If you want this package to have a consistent result for the end user, I would recommend you revert this PR

Issue #276, opened 2015-11-15
Wrong column offset


albertosantini:
In `5.1.0` there is something weird in column offset.

```
"use strict";

var line = 1

console.log(line);

(function (module) {

}());
```

```
λ eslint foo.js

C:\My\Dev\snippets\angular2-es5-seed\foo.js
  3:13  error  Missing semicolon                   semi
  7:12  error  "module" is defined but never used  no-unused-vars

✖ 2 problems (2 errors, 0 warnings)
```

In the panel `eslint-linter` reports
```
ESLint Error semi Missing semicolon.at line 3 col 12
ESLint Error no-unused-vars "module" is defined but never usedat line 7 col 12
```

Notice `line  3 col 12` vs. `3:13`; while the column of the other error is correct.

steelbrain:
turns out, it's weirdly inaccurate. Works for some rules and not for others
![screenshot](https://cloud.githubusercontent.com/assets/4278113/11170391/fe5dcba8-8b8f-11e5-9bb7-a8ae76742550.png)

![screenshot](https://cloud.githubusercontent.com/assets/4278113/11170394/143de200-8b90-11e5-83e0-ffb4b4f0725c.png)

albertosantini:
It seems the problem is only with the issues where the column error is the last char of the line.
I am preparing a PR. 

However it should be better handled in `rangeFromLineNumber`, because the logic is quite weak there, in my humble opionion.

steelbrain:
I would love to accept your contributions to not just this project but to the other one as well. Nothing makes me happier than seeing people collaborate on projects, but it would really help me understand the matter if you would explain what you mean by weak logic :)

albertosantini:
A few notes about `rangeFromLineNumber`. 

If column parameter is not passed, the end column is zero.
This value is not obvious in the code (and there is not documentation).

The column value is zero-based, so it is subtracted one to the value when the range is created,
but the linters use a logic one-based. Indeed they take the start column and add one, for instance, when they need to display info in the panel.

So for weak, I mean the render logic (zero-based, I suppose for underline tasks) is mixed with the editor one (one-based, column starts from one in the status bar). 

Of course, I may be very wrong and it is only an opinion. :)

Arcanemagus:
@albertosantini the `colStart` parameter has nothing to do with the end column, that is derived from `textEditor.getBuffer().lineLengthForRow(lineNumber)`.

[...]

PR #277, opened 2015-11-15
Fix wrong column offset

albertosantini:
When the column error is the last char of the line, due to `rangeFromLineNumber`, the range is malformed: the start column is greater than the end column.

(Also const values cannot be modified) :)

See #276.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -99,7 +99,10 @@ export default {
           nodePath: atom.config.get('linter-eslint.globalNodePath')
         }).then(function(response) {
           return response.map(function({message, line, severity, ruleId, column}) {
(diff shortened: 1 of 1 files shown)


steelbrain:
The patch looks good to me except that you are converting a `const` into a `let`. The code was perfectly fine because we weren't overwriting the variable like
```js
// bad
const range = [1,2,3]
range = [0]
```
but we were instead modifying one of it's properties which is perfectly fine according to the specification
```js
// good
const range = [1,2,3]
range[0] = 2
```

albertosantini:
Ah yes, you are right!

I change it again to `const`.

steelbrain:
LGTM Thanks :+1:

PR #297, opened 2015-11-17
Interpolate env var in conf path

### LINKED ISSUE #291: EsLint of standalone JS not working after v3.1.1
I open JS files in isolation sometimes, when there is no related project or .eslintrc file.

Under these conditions under 3.1.1 it does run EsLint with the default rules, which is nice for catching some issues when scanning code.

This hasn't been working for a while, despite the following in my config.cson.
```javascript
  "linter-eslint":
    disableWhenNoEslintrcFileInPath: false
```
Downgrading to 3.1.1 gives the desired behavior, any version after does not work.

I was thinking someone must have reported this, so have been manually downgrading to 3.1.1 on every update for a few weeks :scream:

SpainTrain:
* :new: Interpolate environment variables in the provided path for the eslint config file

Closes #291

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/297)

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -6,6 +6,7 @@ const CP = require('childprocess-promise')
 const ChildProcess = require('child_process')
 const Path = require('path')
 const FS = require('fs')
+const resolveEnv = require('resolve-env')
 
 const Communication = new CP()
 
@@ -92,7 +93,7 @@ Communication.on('JOB', function(job) {
       Path.join(__dirname, 'reporter.js')
     ]
     if (params.rulesDir) {
-      let rulesDir = params.rulesDir
+      let rulesDir = resolveEnv(params.rulesDir)
       if (!Path.isAbsolute(rulesDir)) {
         rulesDir = find(params.fileDir, rulesDir)
       }
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "atom-package-deps": "^3.0.2",
     "childprocess-promise": "^3.0.0",
     "escape-html": "^1.0.3",
-    "eslint": "^1.9.0"
+    "eslint": "^1.9.0",
+    "resolve-env": "^1.0.0"
   },
   "devDependencies": {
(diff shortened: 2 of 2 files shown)


steelbrain:
While what you did looks nice but I wouldn't trust a module who is not popular and we don't even know it's author, what if he doesn't follow semver and breaks the API?

I would highly recommend we pin the version of it or find an alternative package that is more trust worthy

JohnMurga:
The entire code for the module appears to be 15 lines :-)
https://github.com/LinusU/node-resolve-env/blob/master/index.js

SpainTrain:
I really doubt that will be an issue given that the author has many packages (https://www.npmjs.com/~linusu) and is a member on @expressjs (https://github.com/LinusU)

@LinusU, can you comment on following semver in your projects?

SpainTrain:
RE: LOC - https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328

JohnMurga:
@SpainTrain ... Agreed ... I was thinking more about risk in relation to complexity

SpainTrain:
oh gotcha! :+1:

LinusU:
I promise to follow semver, which I do with all of my modules :)

I also fullheartedly agree with https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328 and I'm usually very fast at responding to bug reports.

SpainTrain:
Thanks for responding!

strangers-npm


@steelbrain are we good to merge?

LinusU:
Awesome :+1:

steelbrain:
@LinusU Thank you for a nice package! :+1:

Arcanemagus:
Published in v5.2.1.

JohnMurga:
Thanks all :+1: 
However, has to extend this to the "eslintrcPath" with this https://github.com/AtomLinter/linter-eslint/pull/305 tiny patch.
Now it works for me in combination with setting the the disableWhenNoEslintrcFileInPath config.

PR #306, opened 2015-11-19
Use relative file paths if possible

IanVS:
`.eslintignore` files must be placed at a project's root, and the patterns within the .eslintignore are relative to project root.  Therefore, for ESLint to know whether a file should be ignored, one of two things must be true:

1. The .eslintignore entries need to be in the format `**/foo.js`
2. The paths passed to eslint for `--stdin-filename` need to be relative to the project root, which is where the .eslintignore must be located.

This change will check for the existence of a `.eslintignore` file, and if one is found, the arguments to `--stdin-filename` will be relative to its directory.

It also adds an `.editorconfig` file to the repo, to make it easier for contributors to use the correct style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -11,7 +11,7 @@ const resolveEnv = require('resolve-env')
 const Communication = new CP()
 
 let eslintPath = null
-let eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), 'node_modules', 'eslint')
+const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), 'node_modules', 'eslint')
 let eslint = null
 let prefixPath = null
 
@@ -42,7 +42,8 @@ Communication.on('JOB', function(job) {
     configFile = params.configFile
   }
 
-  let modulesPath = find(params.fileDir, 'node_modules')
+  const modulesPath = find(params.fileDir, 'node_modules')
+  const eslintignoreDir = Path.dirname(find(params.fileDir, '.eslintignore'))
   let eslintNewPath = null
   if (modulesPath) {
     process.env.NODE_PATH = modulesPath
@@ -85,6 +86,7 @@ Communication.on('JOB', function(job) {
   }
 
   job.Response = new Promise(function(resolve) {
+    const filePath = (eslintignoreDir) ? Path.relative(eslintignoreDir, params.filePath) : params.filePath
     const argv = [
       process.execPath,
       eslintPath,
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+end_of_line = lf
```
Arcanemagus: This is better handled with a `.gitattributes` file, since that is enforced no matter what while this is only relevant for editors that have been modified to handle it.

IanVS: That might make sense for line endings, but as far as I know, `.gitattributes` will not automatically make your editor use two spaces versus four when you're typing, which `.editorconfig` will do.  I see this merely a convenience.   To enforce conventions like spacing, this repo really should be being linted and tested.

Arcanemagus: Hence why that comment was specifically on the `end_of_line` line :wink:.

Arcanemagus:
Is this actually required? As far as I can see this is already working fine, can you provide an example where it isn't?

.editorconfig
Arcanemagus: Yes, this repo absolutely needs to be running CI testing, I'm a bit amazed it hasn't already been set up for that.

lib/worker.js
```diff
@@ -42,7 +42,8 @@ Communication.on('JOB', function(job) {
     configFile = params.configFile
   }
 
-  let modulesPath = find(params.fileDir, 'node_modules')
+  const modulesPath = find(params.fileDir, 'node_modules')
+  const eslintignoreDir = Path.dirname(find(params.fileDir, '.eslintignore'))
```
steelbrain: IMO, we should presume that `.eslintrc` and `.eslintignore` are in the same dir, How does the cli `eslint` handle it? Does it search for both?

lib/worker.js
```diff
@@ -85,6 +86,7 @@ Communication.on('JOB', function(job) {
   }
 
   job.Response = new Promise(function(resolve) {
+    const filePath = (eslintignoreDir) ? Path.relative(eslintignoreDir, params.filePath) : params.filePath
```
steelbrain: We could supply a relative path by default, I don't think it breaks anything :)

steelbrain:
@Arcanemagus This is required to fix a bug where `.eslintignore` doesn't work if the paths in it are relative

lib/worker.js
IanVS: No, they are often not in the same dir. .eslintrc files cascade and are merged, while eslintignore files must be in the project root. See http://eslint.org/docs/user-guide/configuring.html#configuration-cascading-and-hierarchy for more details. 

Besides, the goal here is to provide file paths that can be matched against the patterns in the eslintignore, so it makes sense to do it this way I think.

lib/worker.js
steelbrain: Nevermind, disregard, misread the diff

lib/worker.js
steelbrain: Agreed

lib/worker.js
IanVS: If there is no eslintignore, what would it be relative to? I can't think of any other good way to determine project root. And I don't think absolute paths hurt anything either.

steelbrain:
LGTM, merge at will

Arcanemagus:
Odd, it was working for me with relative paths in `.eslintignore` already.

IanVS:
Were you using the format `**/foo.js`?  Because that would have worked before, but `foo.js` would not have.

Arcanemagus:
I tried `folder/in/project/foo.js`, but not _just_ `foo.js`.

PR #307, opened 2015-11-19
Make `Linter Eslint: Fix File` to use the same ESLint as when linting

### LINKED ISSUE #300: Run fix command in worker
When I run the command manually eslint fixes my file. But the built in command for atom doesn't work.

Example that works:
./node_modules/eslint/bin/eslint.js --fix ./subfolder/myfile.js

I'm not sure how to debug my atom application to find out why the command isn't working...

IanVS:
Moves the execution of `eslint --fix` to the worker, and uses the same eslint resolution as linting.

Fixes #300


lib/main.js
```diff
@@ -70,14 +71,22 @@ export default {
...
       'linter-eslint:fix-file': function() {
         const textEditor = atom.workspace.getActiveTextEditor()
+        const filePath = textEditor.getPath()
+        const fileDir = Path.dirname(filePath)
+        const params = {
+          fileDir,
+          global: atom.config.get('linter-eslint.useGlobalEslint'),
+          nodePath: atom.config.get('linter-eslint.globalNodePath'),
+        }
+        const eslintDir = findEslintDir(params)
```
steelbrain: I would highly recommend that we do not do this in the main process, the sync file system operations and sync execution is gonna make the editor jam for the user

IanVS: OK, I'll have to study up on how to use a worker. Any tips?

steelbrain: [Here](https://github.com/AtomLinter/linter-eslint/blob/master/lib/main.js#L125-L138)'s how you can give a job to worker, the `request` method will return a promise that will resolve when worker replies, You can either return a promise or a value from the worker, like it's done [here](https://github.com/AtomLinter/linter-eslint/blob/master/lib/worker.js#L32), [here](https://github.com/AtomLinter/linter-eslint/blob/master/lib/worker.js#L39) and [here](https://github.com/AtomLinter/linter-eslint/blob/master/lib/worker.js#L88)

IanVS:
Hm, not sure why CI is failing here.  Please give this a thorough review.  It seems to work when I test locally, but without specs it's hard to know for sure that I haven't borked anything up.

Arcanemagus:
CI is failing currently since `apm test` has no specs to test so it always fails. I'm working on those right now :wink:.

lib/worker.js
```diff
@@ -3,89 +3,52 @@
...
-    configFile = params.configFile
   }
 
-  const modulesPath = find(params.fileDir, 'node_modules')
-  const eslintignoreDir = Path.dirname(find(params.fileDir, '.eslintignore'))
-  let eslintNewPath = null
   if (modulesPath) {
     process.env.NODE_PATH = modulesPath
   } else process.env.NODE_PATH = ''
   require('module').Module._initPaths()
```
IanVS: ### Question

What does this line do?

steelbrain: `process.env.NODE_PATH` change is not recognized by the node module loader, this line makes sure it does

IanVS:
@steelbrain & @Arcanemagus, sorry for the rather large PR here.  This does several things:

- Extracts shared/reusable logic into helper functions
- Adds unit tests for those helper functions
- Reorganizes the test fixture files to be more isolated
- Fixes a few bugs I came across while testing
- Finally, fixes #300 and improves the `fix` behavior by alerting the user as to the results.

Please let me know if you'd like to see more tests, or if this needs to be broken up.

lib/main.js
```diff
@@ -73,20 +72,31 @@ export default {
...
-        const fileDir = Path.dirname(filePath)
-        new BufferedNodeProcess({
-          command: Path.normalize(Path.join(__dirname, '..', 'node_modules', 'eslint', 'bin', 'eslint.js')),
-          args: [filePath, '--fix'],
-          options: {
-            cwd: fileDir
-          }
+
+        if (this.worker === null) {
+          return;
```
steelbrain: No semi-colons please :stuck_out_tongue:

lib/es5-helpers.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
+
+const ChildProcess = require('child_process')
+const Path = require('path')
+const FS = require('fs')
+
+let prefixPath = null
+
+function find(startDir, names) {
```
steelbrain: I just published `3.4.0` of helper modules, it's written in es6+ and contains dist files, so it should load in worker now. Means we can use `atom-linter::findFile` (also finds folders) instead of this

lib/main.js
IanVS: It's a hard habit to break. :)  We should turn on `no-extra-semi`.  In fact, I'll do that now and fix any errors.

lib/es5-helpers.js
steelbrain: I am renaming it to `find` in the next version :stuck_out_tongue:

IanVS: Sounds good.  What is the required action here then?  Do you want to just commit whatever changes you want onto this branch, since I'm not sure how the helper should be used?

IanVS: Your helper might want to consider doing the same as: https://github.com/AtomLinter/linter-eslint/commit/4a9387f66ab6896a9525690cb9a3f4494831ae2b.  It fixes a bug I was seeing in the tests where the final path checked ended up being `.` instead of `/`.

lib/main.js
Arcanemagus: Personally I like having them in there, but can live without them. Just as long as the files are consistent :stuck_out_tongue:.

lib/main.js
```diff
@@ -73,20 +72,31 @@ export default {
...
-        const fileDir = Path.dirname(filePath)
-        new BufferedNodeProcess({
-          command: Path.normalize(Path.join(__dirname, '..', 'node_modules', 'eslint', 'bin', 'eslint.js')),
-          args: [filePath, '--fix'],
-          options: {
-            cwd: fileDir
-          }
+
+        if (this.worker === null) {
+          return
```
Arcanemagus: Shouldn't we have a warning that the worker isn't ready to handle the fix request yet?

Arcanemagus:
LGTM other than the null fix worker noted above.

lib/main.js
IanVS: It's really hard to run the `fix` command before 5 seconds has elapsed, but yes, that's a good idea.  I've added the error message.

lib/es5-helpers.js
```diff
@@ -0,0 +1,80 @@
...
+
+const ChildProcess = require('child_process')
+const Path = require('path')
+const FS = require('fs')
+const find = require('atom-linter').findFile
+
+let prefixPath = null
+
+function findEslintDir(params) {
+  const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), 'node_modules', 'eslint')
```
steelbrain: Because the value of this doesn't change and syscalls are expensive, can we move this in upper scope please?

steelbrain:
:clap:

PR #312, opened 2015-11-20
Add specs

### LINKED ISSUE #176: Add test suite
There has been quite a lot of user pain felt over the last several releases because of regressions and new bugs introduced from at least one rewrite. Would you mind adding a test suite for this package? As a user it can be extremely confusing and frustrating when updates seem to break more things than they resolve. Thanks!

Arcanemagus:
~~This does not work as of yet, see comments in the code. I would love some help with this as I haven't been able to get it working as of yet.~~

TODO:
- [x] `.eslintignore` for the CI Builds, but at the same time keeping `bad.js` available for linting for the specs
- [x] Waiting on the worker thread to spawn before running the specs

This PR ended up doing quite a lot to get the specs finished including:
* Fixing #315
* Implementing a flag controlling `--no-ignore`
* Add a message displayed when the worker process is not available

Fixes #176.


Arcanemagus:
Moved this over to the main repo, if anyone has ideas on how to fix this feel free to commit and merge if it's fully working.

Watch out though! Jasmine seems to silently catch thrown errors so you need to watch the console to make sure the errors about `Pane has been destroyed` are not showing up, and I would recommend setting a breakpoint inside each test when you think you have it working to ensure that all the tests are actually being ran as I saw many times 6/6 "working" when in reality only two tests were actually running.

lib/main.js
```diff
@@ -123,7 +123,12 @@ export default {
         const showRule = atom.config.get('linter-eslint.showRuleIdInMessage')
 
         if (this.worker === null) {
-          return []
+          return Promise.resolve([{
```
steelbrain: You don't necessarily have to return a promise, You can just return an array

Arcanemagus: The specs expect a promise to be returned and were erroring without this.

Since this isn't within a `then` it is not automatically a promise.

Arcanemagus:
The specs should have their issues worked out now, we just need a way to force the worker to be enabled and we should be good other than the `.eslintignore` issue that will cause CI to puke over a file not passing the lint.

Arcanemagus:
Implemented a configuration option to ignore any `.eslintignore` files, then used that in the specs to override the one allowing the package to pass linting in CI.

Arcanemagus:
Jasmine is eating errors again, only 2 asserts means it's not running all the specs correctly, investigating...

Arcanemagus:
`4 tests, 13 assertions, 0 failures, 0 skipped`

Good to go!

spec/linter-eslint-spec.js
```diff
@@ -0,0 +1,68 @@
...
+        atom.config.set('linter-eslint.disableEslintIgnore', true);
+        return atom.workspace.open(__dirname + '/files/bad.js').then(openEditor => {
+          editor = openEditor;
+        });
+      });
+    });
+
+    it('finds at least one message', () => {
+      return lint(editor).then(messages => {
+        expect(messages.length).toEqual(1);
```
IanVS: Should it be at least one or exactly one? The assertion doesn't match the description.

spec/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "globals": {
+    "waitsForPromise": true
+  },
+  "env": {
+    "jasmine": true
+  }
+}
```
IanVS: I would suggest creating a few separate subdirectories inside spec/files, and putting the appropriate eslintrc files within these directories, each with "root:true" in them. Or, I can move things around in a separate PR.

IanVS:
> Implementing a flag controlling --no-ignore

It seems to me like this is a separate item from adding tests, and deserves it's own PR. 

Other than that and aa few minor comments I made, this looks :+1:. Awesome job getting it all set up, this will be great to have.

spec/linter-eslint-spec.js
Arcanemagus: Whoops, that got missed in one of the many re-writes trying to get it working.

Arcanemagus:
Yea implementing that really should have been it's own PR, but it was necessary to get the tests working. It really should have been its own PR but I didn't feel like dealing with the mess of splitting it out.

spec/linter-eslint-spec.js
Arcanemagus: I said "at least one" to leave room for future rules giving more errors, currently there is only the one.

PR #313, opened 2015-11-20
Find all node_modules directories

### LINKED ISSUE #294: error with extends local config having sublevels of node_modules folders

```
node_modules/
src/
    components/
    styles/
    node_modules/
    index.js
.eslintrc
gulpfile.js
package.json
```

aruberto:
Don't stop searching for node_modules directories after first directory is found.

Fixes #294.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -42,9 +42,15 @@ Communication.on('JOB', function(job) {
     configFile = params.configFile
   }
 
-  const modulesPath = find(params.fileDir, 'node_modules')
+  const packageJsonPath = find(params.fileDir, 'package.json')
   const eslintignoreDir = Path.dirname(find(params.fileDir, '.eslintignore'))
   let eslintNewPath = null
+  let modulesPath = null
+
+  if (packageJsonPath !== null) {
+    modulesPath = Path.join(Path.dirname(packageJsonPath), 'node_modules')
+  }
+
   if (modulesPath) {
     process.env.NODE_PATH = modulesPath
   } else process.env.NODE_PATH = ''


lib/worker.js
```diff
@@ -34,7 +44,7 @@ Communication.on('JOB', function(job) {
   let configFile = null
   global.__LINTER_RESPONSE = []
 
-  configFile = find(params.fileDir, '.eslintrc')
+  configFile = find(params.fileDir, '.eslintrc', 1)
```
Arcanemagus: Are you sure the config file supports multiple files as a parameter? The wording in [the documentation](http://eslint.org/docs/user-guide/command-line-interface) seems to indicate this should be a single file.

lib/worker.js
```diff
@@ -15,18 +15,28 @@ const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), '
...
+      if (limit && filePaths.length >= limit) {
+        break
+      }
     } catch (_) { }
+
     chunks.pop()
   }
-  return null
+
+  // Return null when nothing found to keep previous find() behaviour
```
Arcanemagus: Only nitpick I have would be to remove this comment, there isn't really a need for it in the code since somebody would have to dig through the history to see what it meant anyway :wink:.

Arcanemagus:
Just tested this with the CLI `eslint` and it doesn't quite match the behavior this PR would change it to. `eslint` from the command line _only_ uses the `node_modules` folder at the same level as the nearest `package.json`, not all `node_modules` folders it finds on it's way up to that file.

We want to match the cli `eslint` as closely as possible, so can you change this PR to match that?

Arcanemagus:
(Edited your comment so the related issue will be automatically closed when this is merged.)

aruberto:
@Arcanemagus Thanks for looking into what eslint CLI does. I can change the PR to search for nearest package.json and use node_modules in that folder instead.

aruberto:
PR has been updated.

lib/worker.js
```diff
@@ -42,9 +42,15 @@ Communication.on('JOB', function(job) {
...
   }
 
-  const modulesPath = find(params.fileDir, 'node_modules')
+  const packageJsonPath = find(params.fileDir, 'package.json')
   const eslintignoreDir = Path.dirname(find(params.fileDir, '.eslintignore'))
   let eslintNewPath = null
+  let modulesPath = null
+
+  if (packageJsonPath !== null) {
+    modulesPath = Path.join(Path.dirname(packageJsonPath), 'node_modules')
```
Arcanemagus: What if there is no such folder, does this cause issues later on when it attempts to use it?

IanVS:
I'd really like to pull out all this nasty manual resolution going on in here, and use something like [npm-which](https://github.com/timoxley/npm-which) instead.  Otherwise we are always going to struggle with weird edge cases like this.

Arcanemagus:
@IanVS should we verify this works and merge it or would you prefer a different solution?

IanVS:
I would prefer we wait and try to do it the right way. I'll take a shot at it tonight.

ide:
@IanVS: npm-which is a great idea. Module resolution is probably the only fragile part of linter-eslint I've encountered so I really like the idea of making it stable.

Were you able to get around to trying npm-which? Or @aruberto, maybe you're able to give it a try?

steelbrain:
Just a side note that `npm-which` spawns an npm process for each call, too expensive, we should only use it when we have evaluated every other option and find them unusable

ide:
@steelbrain: I'd like to profile it since it could be fast enough. O(100ms) when opening a file for the first time to set up the linter in the background would be pretty reasonable to me.

steelbrain:
@ide Not just the first time, it would be for each different file that is linted regardless of first or last time

IanVS:
@steelbrain Are you sure about npm-which spawning an npm process?  I'm talking about https://github.com/timoxley/npm-which, which as far as I can tell does not have npm anywhere in its dependency tree.

steelbrain:
@IanVS I must have looked at the wrong module

Arcanemagus:
@IanVS were you still planning on looking into this? I can take a stab at it if you prefer.

IanVS:
Feel free to give it a shot if you've got some time.  I've got a pretty lengthy backlog right now.

Arcanemagus:
@aruberto you are of course welcome to update this PR, or I'll try to take a stab at it later today.

Arcanemagus:
So a "problem" with using `npm-which` is that if the user has `eslint` installed globally but not locally it will find that and use it even if "Use Global" is unchecked. I'm not sure if this is really an issue though...

ide:
@Arcanemagus The local copy of eslint should take precedence over the global one. This way you can check out other people's repos, or your team's repo at a company, and use the same eslint setup as everyone else working on the same project.

If `nom-which` doesn't meet our needs, it could be good to write a little custom module of our own. The things I like about that are: it can be independently unit tested and have its own semver, so that linter-eslint's lookup behavior is more likely to remain stable.

Arcanemagus:
The local copy does take precedence of course, it's just a global one will take precedence over the one built into the package.

(Unless the system `npm` binaries are not in a `.bin` folder at the same level as the global `node_modules`)

ide:
Ohh... I didn't realize linter-eslint includes a copy of eslint. Giving the global copy precedence over linter-eslint's is kind of nice actually.

Arcanemagus:
Unfortunately it looks like `npm-which` fails in the same manner as the current behavior before this PR due to a bug in it's PATH generator (https://github.com/timoxley/npm-path/issues/2).

Arcanemagus:
For now I'm going to close this PR, if #350 hasn't been merged yet and you feel like updating this PR @aruberto feel free as it would be nice for you to get credit for the initial work :wink:.

aruberto:
@Arcanemagus Thanks for putting #350 together. No need to worry about credit, the issue getting fixed with the correct solution is more important than me getting credit :smile:

timoxley:
Note that the `npm-path` issue has been fixed, so `npm-which` may be again be a viable solution here.

PR #318, opened 2015-11-21
Check for config in package.json and .eslintrc before disabling

### LINKED ISSUE #226: rename/update disableWhenNoEslintrcFileInPath to disableWhenNoEslintConfig
eslint config can be in the package.json under an "eslintConfig" field.
That could be a nice enhancement.

tschaub:
This adds a `disableWhenNoEslintConfig` setting.  When `true` (the default), the linter will be disabled only if no config is found in a `package.json` or `.eslintrc` file.

This removes the `disableWhenNoEslintrcFileInPath` setting (which breaks the use of `eslintConfig` in `package.json`).

I couldn't see an easy way to add tests for this.  Let me know if you have any suggestions.  In addition, I'd be happy to rework this if you would prefer to reuse the original setting (or keep it and add the new one).

Fixes #226.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -32,10 +32,17 @@ function find(startDir, name) {
 Communication.on('JOB', function(job) {
   const params = job.Message
   let configFile = null
+  let configInPackage = false
   global.__LINTER_RESPONSE = []
 
   configFile = find(params.fileDir, '.eslintrc')
-  if (params.canDisable && configFile === null) {
+  if (!configFile) {
+    const packagePath = find(params.fileDir, 'package.json')
+    if (packagePath) {
+      configInPackage = Boolean(require(packagePath).eslintConfig)
+    }
+  }
+  if (params.canDisable && !configFile && !configInPackage) {
     job.Response = []
     return
   } else if (params.configFile) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ### Upcoming
 
 * Allow ignoring the `.eslintignore` file
+* Add `disableWhenNoEslintConfig` config (`true` by default) to only disable the linter if no ESLint config is found in a `package.json`
+or `.eslintrc` file.  This replaces the `disableWhenNoEslintrcFileInPath` config.
 
 ### v5.2.1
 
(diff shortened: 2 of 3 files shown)


lib/worker.js
```diff
@@ -32,10 +32,17 @@ function find(startDir, name) {
...
   let configFile = null
+  let configInPackage = false
   global.__LINTER_RESPONSE = []
 
   configFile = find(params.fileDir, '.eslintrc')
-  if (params.canDisable && configFile === null) {
+  if (!configFile) {
+    const packagePath = find(params.fileDir, 'package.json')
+    if (packagePath) {
+      configInPackage = !!require(packagePath).eslintConfig
```
steelbrain: How about `Boolean(require(packagePath).eslintConfig)`

tschaub: Updated in 73be80900626ab0f7b29b081b39511715caf0e0d.

steelbrain:
Now all we need is a change log entry under the `Upcoming` title

Arcanemagus:
I would _much_ rather see an implementation that takes advantage of [`getConfigForFile()`](http://eslint.org/docs/developer-guide/nodejs-api#getconfigforfile). We should be offloading as much work (and thus implementation) onto `eslint` as we can, especially since this means we match `eslint` as closely as possible.

steelbrain:
@Arcanemagus We **could** use that in the future, when we use the cli engine class and all, but that's not what we're using right now so it won't work for us. I'd love for us to do it in the future.

Till then, this LGTM

IanVS:
@steelbrain, I was wondering about that.  Why *aren't* we using the cliEngine?

Arcanemagus:
@IanVS from the sounds of it: Because nobody has implemented it :stuck_out_tongue:

Sounds like it's time for a new project...

steelbrain:
@IanVS I am not entirely sure, the last version did use it but it was failing in a lot of places, so I just monkey patched a worker process that would work™ as fast as I could, and the end result wasn't beautiful.

I really think we should, time for a rewrite :muscle:

steelbrain:
@IanVS Are you on Atom Slack?

IanVS:
@steelbrain I am now.

PR #319, opened 2015-11-21
Support File Formats

### LINKED ISSUE #317: Support Configuration File Formats
ESLint 1.10 deprecates `.eslintrc` and recommends using `.eslintrc.js`, `.eslint.yaml`, `eslint.yml`, or `.eslint.json`.

Reference http://eslint.org/blog/2015/11/eslint-v1.10.0-released/#configuration-file-formats

ChristianMurphy:
resolves #317

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -15,15 +15,24 @@ const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), '
 let eslint = null
 let prefixPath = null
 
-function find(startDir, name) {
-  let filePath
+function find(startDir, names) {
+  let localNames;
+  if (typeof name === 'string') {
+    localNames = [names]
+  } else {
+    localNames = names
+  }
   const chunks = startDir.split(Path.sep)
   while (chunks.length) {
-    filePath = Path.join(chunks.join(Path.sep), name)
-    try {
-      FS.accessSync(filePath, FS.R_OK)
-      return filePath
-    } catch (_) { }
+    const currentDirectory = Path.join(chunks.join(Path.sep))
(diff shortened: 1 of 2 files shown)


steelbrain:
@ChristianMurphy Be advised, this PR has got merge conflicts

ChristianMurphy:
Resolved conflicts

lib/worker.js
```diff
@@ -15,15 +15,18 @@ const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), '
 let eslint = null
 let prefixPath = null
 
-function find(startDir, name) {
-  let filePath
+function find(startDir, names) {
```
steelbrain: This function is used in several places, not all of them pass array, so you should add something like
```js
if (typeof names === 'string') {
  names = [names]
}
```

ChristianMurphy: Updated to test for type `string`.
Added a new `localNames` variable to avoid param reassign.

steelbrain:
Now all we need is a CHANGELOG entry

ChristianMurphy:
Added CHANGELOG entry

steelbrain:
CircleCI seems to be playing dumb, anyways, LGTM. Thanks!

Arcanemagus:
When it fails at an intermediate step like that you should just restart the build.

lib/worker.js
```diff
@@ -15,15 +15,24 @@ const eslintPathLocal = Path.join(FS.realpathSync(Path.join(__dirname, '..')), '
...
+  let localNames;
+  if (typeof name === 'string') {
```
tschaub: This should be `names`.  See #325 for a proposed fix.

PR #329, opened 2015-11-23
Update deps

Arcanemagus:
Update the dependencies to grab a fix in `atom-linter`. Adds a few badges to the README.


iam4x:
:ship:

PR #336, opened 2015-11-26
Communicate worker start depends on user action

sheeley:
Attempting to keep users from thinking that they need to wait for the `worker not initialized yet` message to go away.


steelbrain:
LGTM, Thanks

PR #340, opened 2015-11-28
Fix file import resolution

### LINKED ISSUE #335: Version 5.2.3 breaks import/no-unresolved rule
I haven't dug into what could be causing this, but when upgrading from `5.2.2` to `5.2.3` I begin to get failures from [import/no-unresolved](https://github.com/benmosher/eslint-plugin-import#no-unresolved) for all of my relative imports.

### 5.2.2

screen shot 2015-11-25 at 9 05 56 am

### 5.2.3

screen shot 2015-11-25 at 9 14 05 am

IanVS:
Fixes #335

I originally didn't realize that `process.cwd()` is always the directory of the file being linted.  Because of this, it's possible to pass in only the basename of the file, and ESLint will assume cwd is set correctly, and will resolve configs and imports.

But, this breaks use of `.eslintignore` files, so if one of those are found, we can set cwd to the project root and pass the relative path to eslint.


lib/worker.js
```diff
@@ -102,7 +102,13 @@ Communication.on('JOB', function(job) {
   }
 
   job.Response = new Promise(function(resolve) {
-    const filePath = (eslintignoreDir) ? Path.relative(eslintignoreDir, params.filePath) : params.filePath
+    let filePath
+    if (eslintignoreDir) {
+      filePath = Path.relative(eslintignoreDir, params.filePath)
+      process.chdir(eslintignoreDir)
+    } else {
+      filePath = Path.basename(params.filePath)
```
steelbrain: Wouldn't it be nice to move [this](https://github.com/AtomLinter/linter-eslint/blob/issue335/lib/worker.js#L69) here?

IanVS: Ah, completely missed that.  Yes, indeed it would.  One moment.

IanVS: ok, moved.

steelbrain:
LGTM, merge if CI passes

IanVS:
I guess AppVeyor isn't working correctly? https://ci.appveyor.com/project/Arcanemagus/linter-eslint/build/50

Arcanemagus:
AppVeyor will be broken till v1.3.0 is released, hence why it isn't marked as required :wink:.

IanVS:
Thanks @Arcanemagus.  This look okay to you?

package.json
```diff
@@ -19,9 +19,10 @@
   },
   "devDependencies": {
     "babel-eslint": "^4.1.5",
-    "eslint-plugin-react": "^3.10.0",
+    "eslint-config-airbnb": "latest",
```
Arcanemagus: Alphabetising them? :stuck_out_tongue:

IanVS: `npm i --save-dev` does this automatically, apparently.

Arcanemagus:
Looks good to me :+1:.

PR #342, opened 2015-11-30
Support for multiple rules dirs

christopherstott:
I added support for passing in multiple rules dirs to eslint.

Added a new array config property eslintRulesDirs (left eslintRulesDir as is to not break existing functionality).

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/342)


steelbrain:
Instead of having two configs `eslintRulesDirs` and `eslintRulesDir`, how about we modify `eslintRulesDir` to accept an array and keep that key for backward compatibility and add a custom title with `Dirs`?

christopherstott:
That's what I was originally hoping to do.

Is there a way to do that and pass the type validation? Currently 'eslintRulesDir' must be a 'string'.

I'm not sure what you mean by "add a custom title with Dirs" ?

Thanks

steelbrain:
```js
    eslintRulesDir: {
      title: 'ESLint Rules Dirs',
      description: 'Specify comma separated directories for ESLint to load rules from',
      type: 'array',
      default: []
    }
```
The key remainds same `eslintRulesDir` for backward compatibility and now we accept an array

Arcanemagus:
What about existing configurations @steelbrain?

steelbrain:
@Arcanemagus If we do what I suggested, it should be backward compatible

Arcanemagus:
So it would read in any existing settings without issue?

steelbrain:
@Arcanemagus True

Arcanemagus:
@christopherstott Have you had a chance to update your PR yet?

Arcanemagus:
@christopherstott Can you rebase this PR and address @steelbrain's comments?

Arcanemagus:
Marking this as closed due to a lack of response.

PR #350, opened 2015-12-04
Fix node_module usage

### LINKED ISSUE #294: error with extends local config having sublevels of node_modules folders

Arcanemagus:
Re-implements #313 as we haven't heard from @aruberto in a while.

Uses the node_module at the level of the nearest `package.json` if one exists, otherwise it will search for one starting at the current directory and moving up.

Fixes #294.

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/350)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
 node_modules
 *.log
 .idea
-
-# Keep node_modules test fixtures
-!spec/fixtures/local-eslint/node_modules
--- a/lib/es5-helpers.js
+++ b/lib/es5-helpers.js
@@ -9,7 +9,13 @@ let prefixPath = null
 const atomEslintPath = Path.join(FS.realpathSync(Path.join(__dirname, '..')), 'node_modules', 'eslint')
 
 function findEslintDir(params) {
-  const modulesPath = find(params.fileDir, 'node_modules')
+  let modulesPath = null
+  const packageJsonPath = find(params.fileDir, 'package.json')
+  if (packageJsonPath !== null) {
+    modulesPath = Path.join(Path.dirname(packageJsonPath), 'node_modules')
+  } else {
+    modulesPath = find(params.fileDir, 'node_modules')
+  }
   let eslintNewPath = null
 
   if (params.global) {
--- a/spec/fixtures/local-eslint/.gitignore
+++ b/spec/fixtures/local-eslint/.gitignore
@@ -1,0 +1,2 @@
+# Keep node_modules test fixtures
+!node_modules
--- a/spec/fixtures/local-eslint/lib/node_modules/.gitignore
+++ b/spec/fixtures/local-eslint/lib/node_modules/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore


Arcanemagus:
@IanVS if you have a better way of doing this feel free to update this branch, `npm-which` unfortunately currently isn't a viable solution since it fails in the same situation our current search fails in due to https://github.com/timoxley/npm-path/issues/2.

Arcanemagus:
@IanVS any comments?

grabbou:
I would love to see that merged as I have a single repository for multiple npm projects and I am keeping `eslint` at the top to have unified linting across packages. Unfortunately, I am getting errors about either missing eslint or eslint-plugin-babel - I would really love that error to be gone.

grabbou:
Is it possible for this module to look up the tree if `eslint` is not present in the current node_modules? I mean - I have few nested `package.json` projects inside a bigger one and wanted to keep `eslint` `eslint-babel` and all that stuff in the umbrella top-level one only. Unfortunately, it looks like it does not look up the tree but fails w/o `eslint` present - esp. given that's how `require` works.

Arcanemagus:
@grabbou I think that is a bit beyond what we are trying to accomplish here. The goal of this is to properly load the `eslint` _local_ to the current project, and if one isn't found to fall back on the global installation if enabled, or the bundled one if not.

This PR handles the case outlined in #294 where there is a node_modules folder within a folder, but that isn't the root of the project.

For your use case I'm not sure how we would go about handling that, search for a `package.json`, check if there is an `eslint` installed at that level and if not go up a directory and start over? We would no longer be at the "local" level for the project though. It wouldn't be strictly correct, but I wouldn't be opposed to also searching at the root of what Atom considers to be the project. Thoughts @AtomLinter/linter-eslint?

grabbou:
>  but I wouldn't be opposed to also searching at the root of what Atom considers to be the project.
+1 to the root folder in the left sidebar.

I totally understand reasoning behind that, just wondering what were the cons of using `require.resolve` if it can save us a couple of LOCs and unit tests (and giving us pretty cool require mechanism out of the box) ?

Arcanemagus:
The directory structure in #294 doesn't have `package.json` files at the sub levels though :wink: Something to do with browserify.

As for why we don't just use `require.resolve`: I didn't think of it, will look into the behavior of that now.

grabbou:
Awesome! Thank you :) From my internal testing and taking look at the source code, it looks like it could save you a couple of `find` calls - tried to call it from a couple of nested folders and it been reporting location w/o issues.

The only case you have to be aware of is that it will return global path if no local installation is found - so in case you don't want to do it by default but leave it behind a flag - you have to probably `indexOf` on path to check if it's global and warn/error in that case with something like `There's no eslint installed locally, but we have found global installation. Turn on `global` flag to skip that warning` or something.

grabbou:
PS. Here's the nice step-by-step break down just in case https://nodejs.org/api/modules.html#modules_all_together

grabbou:
Just to put some more details on that issue -> my repo has exactly the same structure as `babel` - https://github.com/babel/babel and I am afraid `linter-eslint` wont work with it either because of the fact that eslint and babel-eslint are defined in the outer `umbrella` package as well.

https://github.com/babel/babel/blob/master/package.json#L16

Arcanemagus:
I'm probably not going to have time to look into this till later tonight, if you want to feel free to work off of this branch and submit a PR with any changes.

grabbou:
No worries, I'll take a look in few hours and submit PR to this PR with proposed changes so it's easier to discuss :) Thanks for looking into that!

moretti:
@grabbou Have you found any workaround for this issue?

I'm in the exact situation, my project has the same structure as babel, I can run eslint from the CLI, but `atom-eslint` is completely broken.
I've temporary switched to VSCode + `vscode-eslint`, which works fine even in this particular case.

grabbou:
Well, maybe I am missing the point entirely here, but to me, that entire function https://github.com/AtomLinter/linter-eslint/blob/pr313-lda/lib/es5-helpers.js#L11 could be just replaced with `require.resolve('eslint')`.

Arcanemagus:
Closing in favor of #396.

PR #354, opened 2015-12-06
Fix exception for "Fix File" on Windows

### LINKED ISSUE #353: 5.2.5 - Fix File - Linting Errors Remain
On Windows 10, using `Fix File` command with `5.2.5` I get always the following message:
`Linter-ESLint: Fix Attempt Completed, Linting Errors Remain`

Settings:
```
 "linter-eslint":
    useGlobalEslint: true
```

Investigating it seems the config file is not correct, like `C:\my\project\path.eslintrc.yml`: notice missing backslash between path and the config file.

https://github.com/AtomLinter/linter-eslint/blob/master/lib/worker.js#L88

I cannot give more details because, at the moment, for every change I get the annoying spawn bug.

albertosantini:
This patch fixes the exception raised applying "Fix File" command.

It doesn't follow the disruptive suggestion to assimilate on('FIX' to on('JOB' , because there are too much IF statements.
Maybe we can refactor the listeners later and in another way.

Closes #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Upcoming
+
+* Fix file exception on Windows resolution (#353)
+
 ### v5.2.5
 
 * Fix file import resolution (#340)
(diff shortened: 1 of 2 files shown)


steelbrain:
LGTM

Arcanemagus:
LGTM as well, merging and thanks!

PR #367, opened 2015-12-11
WIP: Rewrite in ES6

### LINKED ISSUE #366: Convert to ES6
Okay guys, I am gonna do an iteration 2 of converting the package to es6. Here's how I am gonna do it:

1. I'm gonna create a branch named es6 based on master
2. I'm gonna send PRs TO es6
3. Eventually, gonna merge that es6 into master

I'm gonna do smaller, focused PRs to make review easier

__Note:__ APM uses the repo releases tags from github to download packages, so we can't `.gitignore` dist files, we'll have to include them in the repo

Here are some things that I would try to solve while doing this:

- [ ] A spec for everything
- [x] Unify both worker jobs in one to avoid code de-duping and getting out of sync
- [x] Unify both helper modules

cc @SpainTrain @Arcanemagus @IanVS @iam4x @Adezandee

steelbrain:
:rotating_light: Do not merge unless all of the child PRs are merged :rotating_light: 

Fixes #366

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/367)


albertosantini:
I tested also this branch with atom 1.4.0: indeed compact branch has been merged in this one.

It works ok with my workflow.

The only glitch I noticed it is with `Fix file`:
`Uncaught (in promise) Error: Notification must be created with string message: Error: EPERM: operation not permitted, write(…)`

in `...\atom\app-1.4.0\resources\app.asar\src\notification.js:25`.

(Give, please, a look at timecop, if there is any difference. With compact branch the activation time was very high; with this one it is ok)

steelbrain:
@albertosantini Fixed by https://github.com/AtomLinter/linter-eslint/commit/a817716d49d3c5b54720c5c6660fe945fbce1d68

steelbrain:
Found a few problems, fixing

steelbrain:
@albertosantini Fix file is now completely stable

steelbrain:
Fixing specs now

albertosantini:
LGTM! :) 

Also timecop is ok. 

Thanks for the effort.

steelbrain:
AppVeyor's aggressive caching made this fail, passed on both CircleCI and Travis

nmote:
What is the process for contributing now? I'd like to put up a PR for adding autofix (https://github.com/nmote/linter-eslint/commit/d5c8b30a02904840b178c97c72d269ac12f43430) but the new directory structure is confusing me. It seems like `src` is where the source files live but transpiled code is also getting checked in in `lib`? However there is no documentation on how to run the transpilation, and `babel` isn't even listed as under `devDependencies`.

Also, why check in the transpiled files at all? Atom will do it for you if you have `use babel` at the top, which it seems you do. And even if you don't want to use that feature, can't the transpiled files just be generated in a `prepublish` step? Developers would have to run it anyway to see their changes reflected.

steelbrain:
@nmote @Arcanemagus was just writing the docs, till then you can just manually transpile using babel

steelbrain:
@nmote The reason we're transpiling files is that linter-eslint spawns a child process that does the real lnting in a separate thread and there's no `use babel` available to it

Arcanemagus:
Ha, had gotten distracted, will start that PR now!

Arcanemagus:
And the reason the transpiled files are checked in is for some bizarre reason `apm` uses GitHub tags instead of zips of the archive, so the tagged version has to be the complete plugin. We have toyed with the idea of committing them during build, tagging that, then doing a commit that removes them (automated in the CI process) but decided against it due to the crazy pollution of the history that would bring.

nmote:
Ah interesting, thanks for the explanations. As long as it's documented and doesn't require a setup process beyond `npm install` I'll be content. Please CC me in the documentation PR so I can move my autofix PR along ASAP.

nmote:
Although you could always cut a release branch, do the build, and then tag and it wouldn't pollute the history for `master`. That's still an annoying release process though.

Arcanemagus:
@nmote That actually doesn't sound bad at all. Ideally the whole thing would be automated in CI so the whole process would only need to be setup once.

That would leave around branches for each release though...

nmote:
Well, technically I believe you could delete the branch or never make it, and have just the tag point to the released version. But branches are light-weight enough that polluting the namespace is the only real concern here.

Arcanemagus:
I'd much prefer polluting the branch namespace than including the transpiled code in the repo, but since a tag references a specific commit I _think_ we can get away with creating a branch, comitting the transpiled code there, pushing it up, `apm publish` from there, then deleting the branch. Since the commit should still be in the repo history it should still be a valid tag.

Will look into this tonight, unless you want to take a stab at it @SpainTrain?

PR #389, opened 2016-01-01
add info about aribnb preset

rofrol:
[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/389)


Arcanemagus:
Installing the config globally does nothing unless they add it to their project's `eslint` configuration.

I'm not sure this is really the best place to document things like this, this package is only meant to be an intermediary between Linter and `eslint`, detailed documentation of the configuration should be handled by the `eslint` pages, or the Getting Started guide we link already.

PR #396, opened 2016-01-07
Better `eslint` lookup

### LINKED ISSUE #294: error with extends local config having sublevels of node_modules folders

grabbou:
Fixes #294 and my case outlined in #350 as well. It's an improved version of #350 after some discussion, specifically, it does not rely on presence of `package.json` in sub-projects - #350 was assuming they are not defined, which seems to be only applicable to browserify projects.

It essentially works in the same way as lookup for `.eslintrc` - find the closest `node_modules/eslint` folder starting from the `fileDir` and moving up.

The feature is in lines 15-17, the rest is just styling tweaks - early return to make it easier to reason about and better scoping of variables.

Tested in failing projects with `babel-eslint` and `eslint-plugin-react` plugins.

CC: @Arcanemagus - sorry for the delay! Almost forgot about that. Commit(s) to be rebased once approved / discussed / whatever!

To clarify, that PR makes projects following `babel-like` structure work again, that is:
```
./.eslintrc
./node_modules/eslint
./packages/core/node_modules/...
./packages/web/node_modules/...
```
so editing files inside sub-project with its own `node_modules` won't fail, because we will locate them in the umbrella outer package.

See https://github.com/babel/babel/blob/master/package.json#L16 for real life example.

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/396)

### DIFF
--- a/lib/es5-helpers.js
+++ b/lib/es5-helpers.js
@@ -9,35 +9,27 @@ let prefixPath = null
 const parentPath = FS.realpathSync(Path.join(__dirname, '..'))
 const atomEslintPath = Path.join(parentPath, 'node_modules', 'eslint')
 
+// Find directory of eslint installation. Traverse up the tree checking every
+// `node_modules` directory
 function findEslintDir(params) {
-  const modulesPath = find(params.fileDir, 'node_modules')
-  let eslintNewPath = null
+  if (!params.global) {
+    return find(params.fileDir, 'node_modules/eslint') || atomEslintPath
+  }
 
-  if (params.global) {
-    if (params.nodePath === '' && prefixPath === null) {
-      const npmCommand = process.platform === 'win32' ? 'npm.cmd' : 'npm'
-      try {
-        prefixPath = ChildProcess.spawnSync(npmCommand, ['get', 'prefix']).
-          output[1].toString().trim()
-      } catch (e) {
-        throw new Error('Unable to execute `npm get prefix`. ' +
-          'Please make sure Atom is getting $PATH correctly')
-      }
-    }
-    if (process.platform === 'win32') {
-      eslintNewPath = Path.join(params.nodePath || prefixPath, 'node_modules', 'eslint')
-    } else {
-      eslintNewPath = Path.join(params.nodePath || prefixPath, 'lib', 'node_modules', 'eslint')
-    }
-  } else {
+  if (params.nodePath === '' && prefixPath === null) {
+    const npmCommand = process.platform === 'win32' ? 'npm.cmd' : 'npm'
     try {
-      FS.accessSync(eslintNewPath = Path.join(modulesPath, 'eslint'), FS.R_OK)
-    } catch (_) {
-      eslintNewPath = atomEslintPath
+      prefixPath = ChildProcess.spawnSync(npmCommand, ['get', 'prefix']).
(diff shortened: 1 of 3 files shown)


lib/es5-helpers.js
```diff
@@ -9,35 +9,27 @@ let prefixPath = null
 const parentPath = FS.realpathSync(Path.join(__dirname, '..'))
 const atomEslintPath = Path.join(parentPath, 'node_modules', 'eslint')
 
+// Find directory of eslint installation. Traverse up the tree checking every
+// `node_modules` directory
 function findEslintDir(params) {
-  const modulesPath = find(params.fileDir, 'node_modules')
-  let eslintNewPath = null
+  if (!params.global) {
+    return find(params.fileDir, 'node_modules/eslint') || atomEslintPath
```
grabbou: It turns out that it will keep looking up the tree or return null if it does not exist, so that's the old behaviour of `try catch` we had.

steelbrain: I don't think the `find` method supports `node_modules/eslint`, `eslint` would work though

Arcanemagus:
Can you include specs to test this functionality? Specifically they should be failing before, and working after this patch is applied (and of course match what `eslint` itself does in the terminal :stuck_out_tongue:).

Just including the 3 spec files from #350 should cover the use cases it was originally handling.

lib/es5-helpers.js
grabbou: It seemd to work when I was debugging. with `eslint` it won't go inside `node_modules`, would it?

grabbou:
Ahh, all right! I was looking at test cases in your PR but what I've only seen were empty files - now I know their purpose! Updating :)

lib/es5-helpers.js
```diff
@@ -9,35 +9,27 @@ let prefixPath = null
 const parentPath = FS.realpathSync(Path.join(__dirname, '..'))
 const atomEslintPath = Path.join(parentPath, 'node_modules', 'eslint')
 
+// Find directory of eslint installation. Traverse up the tree checking every
+// `node_modules` directory
 function findEslintDir(params) {
-  const modulesPath = find(params.fileDir, 'node_modules')
-  let eslintNewPath = null
+  if (!params.global) {
+    return find(params.fileDir, ['node_modules', 'eslint'].join(Path.sep)) || atomEslintPath
```
grabbou: Looks like it does -> https://github.com/AtomLinter/atom-linter/blob/master/src/helpers.js#L210. Since it splits by Path.sep - rewrote that from `node_modules/eslint` to join(Path.sep) so it's always split correctly into chunks.

grabbou:
Update: Squashed & reordered so there's failing test suite commit before actual implementation.

Arcanemagus:
@AtomLinter/linter-eslint This is looking good to me, anyone else have any comments?

Arcanemagus:
Got confirmation from @steelbrain in chat that this looks good to him, with no other comments I'm merging this in :stuck_out_tongue:.

grabbou:
Tkanks! Hopefully that wasn't my last PR! :D

PR #373, opened 2015-12-11
Cleanup the codebase

steelbrain:
Changes so far
 
- Log stuff to console as recieved instead of on exit
- Remove timeout from worker spawn
- Use module.exports instead of exports because babel 6's es6 default export is incompatible with commonjs

Fixes #364

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/373)

PR #403, opened 2016-01-20
Add A Configurable Timeout For Respawning A Worker

joefitzgerald:
- Prevents 1000 degree temperatures on your lap if ESLint or Atom is misbehaving

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/403)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -114,19 +114,20 @@ export default {
           this.worker = null
           // Respawn if it crashed. See atom/electron#3446
           if (shouldLive) {
-            let respawnAfter = atom.config.get('linter-eslint.respawnWorkerAfter')
             if (this.exitNotification) {
               this.exitNotification.dismiss()
             }
-            this.exitNotification = atom.notifications.addWarning('[Linter-ESLint] Worker died unexpectedly',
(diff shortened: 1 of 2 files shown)


steelbrain:
This is already done in https://github.com/AtomLinter/linter-eslint/pull/373 along with other improvements.

joefitzgerald:
> This is already done in #373 along with other improvements.

OK, but if review of #373 is going to take a while, can we ship this independently?

lib/main.js
```diff
@@ -102,11 +109,25 @@ export default {
         const { child, worker, subscription } = spawnWorker()
         this.worker = worker
         this.subscriptions.add(subscription)
+
         child.on('exit-linter', shouldLive => {
           this.worker = null
           // Respawn if it crashed. See atom/electron#3446
           if (shouldLive) {
-            initializeWorker()
```
steelbrain: Crashes don't happen all the time for every user, IMO instead of having a config and all of that, we could just add a default delay of 5 seconds here and it'll work good

steelbrain:
Fixed by https://github.com/AtomLinter/linter-eslint/pull/367

PR #415, opened 2016-01-26
Implement autofix

nmote:
Implements the autofix API described in https://github.com/atom-community/linter/issues/1049.

Tested manually within Nuclide and also added a test case.

Everything under `lib` was autogenerated when I ran `npm run watch`.

[Review on Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/415)


nmote:
I separated the build artifact updates into a separate commit, which seems a bit cleaner but it also means that the first commit alone will have a test failure. Let me know if you don't like that.

lib/helpers.js
```diff
@@ -57,4 +57,4 @@ function showError(givenMessage) {
     detail: detail,
     dismissable: true
   });
-}
+}
```
Arcanemagus: Until @steelbrain can get a plugin put together for `ucompiler` I think I'm just going to stop fighting this missing newline in the transpiled code :confused:...

src/main.js
```diff
@@ -149,7 +161,8 @@ module.exports = {
             const ret = {
               filePath,
               type: severity === 1 ? 'Warning' : 'Error',
-              range
+              range,
+              fix: linterFix
```
Arcanemagus: Should this only be present when a fix is available?

nmote: Yeah, it probably should. I'll update.

nmote: (It works in Nuclide since we check for presence via `message.fix != null`, but it should probably get left off altogether if there is no fix)

nmote:
Hold on, I am having more GitHub problems. Sorry.

nmote:
Alright, should be good to go. I also added a test to make sure that the `fix` property does not exist if the error does not have a fix on it. The '"foo" is not defined' error is unlikely to ever have an autofix so I think this should be reliable.

Thanks for bearing with me while I struggle with a platform I'm not familiar with.

Arcanemagus:
LGTM, any changes @AtomLinter/linter-eslint?

steelbrain:
LGTM 2

steelbrain:
Thanks @nmote :+1:

Arcanemagus:
Published in v6.1.0. :tada:

PR #475, opened 2016-03-08
add link to rule definition

granteagon:
User can now click the rule to go to the rule definition in their
browser.

Dist files have been updated.

Please note: not because of a change of mine, but the dist files now have all the const and let definitions replaced with vars.  Verified by @Arcanemagus.


Arcanemagus:
Yea... it's quite bizarre. Something seems to have changed with the babel configuration since they were last updated.

PR #478, opened 2016-03-09
Pin eslint version

Arcanemagus:
Pin the eslint version as a few other bugs are being held back in diagnosis due to CI failing. If users need a newer version one locally installed in their project will work.

This can be removed once `babel-eslint` and `eslint@2.3.0` figure out what they are going to do...


steelbrain:
LGTM

Arcanemagus:
Almost 30 minutes later Travis-CI finally decided to run the OS X builds...

PR #481, opened 2016-03-11
Pass ignore file to eslint explicitly

### LINKED ISSUE #448: .eslintignore file has no effect
I'm on version `6.0.0`. I have a directory structure like this:

```
js/
- node_modules/
- lib/
  - gen/
- .eslintignore
- .eslintrc.json
- .gitignore
- package.json
```

My `.eslintignore` looks like this:
```
node_modules/
lib/gen/
```

And my `.gitignore` file looks like this:
```
lib/gen/
```

(I don't .gitignore my node_modules)

In Atom with `linter-eslint@6.0.0`, opening any `.js` files inside my `node_modules` directory will result in half of my screen being bombarded with lint errors. Opening any `.js` files inside the `lib/gen` directory will not. Furthermore, if I remove `lib/gen` from my `.eslintignore` file, the files inside that directory still do not get linted by the text editor. Only when I remove the path from my `.gitignore` does the linter actually activate for files in that directory.

With some investigation, I added a `console.log()` statements above line 149 of `linter-eslint/lib/main.js` and discovered that opening any files that are under `.gitignore` don't ever even trigger the worker job request.

IanVS:
Fixes #448 

I honestly have no idea why ESLint is not finding the `.eslintignore` file on its own sometimes.  Sometimes it does, and sometimes it doesn't.  With these changes, if we find a `.eslintignore` file, we will pass it to ESLint with the `--ignore-path` arg, to be sure it's picked up.

### DIFF
--- a/lib/worker-helpers.js
+++ b/lib/worker-helpers.js
@@ -135,6 +135,11 @@ function getArgv(type, config, filePath, fileDir, givenConfigPath) {
   }
   argv.push('--format', _path2.default.join(__dirname, 'reporter.js'));
 
+  var ignoreFile = config.disableEslintIgnore ? null : (0, _atomLinter.findCached)(fileDir, '.eslintignore');
+  if (ignoreFile) {
+    argv.push('--ignore-path', ignoreFile);
+  }
+
   if (config.eslintRulesDir) {
     var rulesDir = (0, _resolveEnv2.default)(config.eslintRulesDir);
     if (!_path2.default.isAbsolute(rulesDir)) {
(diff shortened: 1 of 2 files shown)


steelbrain:
@IanVS does it make sure it works? Maybe we should add some specs about it

IanVS:
Strangely enough, there is a spec which was failing before my change, and passing after it.  Not sure why CI didn't pick up on it though.  Is there a difference when running specs from within Atom (`Run Package Specs`)?

Arcanemagus:
There shouldn't be, but it seems some environments that spec works in, and some it doesn't (It was failing for me too on my workstation, but not my laptop...).

Arcanemagus:
This LGTM btw, did you have anything else you were waiting on?

IanVS:
No, I think this is ready to go.

PR #513, opened 2016-03-28
Allow ~/ in path to ESLint Rules Dir (fixes #441)

### LINKED ISSUE #441: linter-eslint breaks when using ~ in the rule path
When setting the "ESLint Rules Dir" in the settings for linter-eslint, if you specify your path with a "~" character, the linter will completely stop working. No errors messages will appear, no warnings will appears, it will simply behave as if it's not installed.

Example of broken path:

```
~/Git/myproject/rules
```

However, if you change that path to an absolute path:

```
/home/myuser/Git/myproject/rules
```

Everything will work correctly again.

If ~ isn't supported in the paths, the linter should at least give us an error message. But better yet -- it would be great to fix this bug.

kaicataldo:
Took a crack at this - this works locally for me on a machine with OSX, but not currently able to test on a Windows or Linux machine. Not sure what is expected/necessary regarding tests. I didn't see any coverage for this particular function, but am happy to write some tests if need be.

The problem seems to be that `path.isAbsolute('~/path/to/rulesConfig')` returns `false`, and then the linter tries to resolve the path relative to `fileDir`. As a result, `rulesDir` has the value `null`, and it silently fails.

I actually can't seem to figure out how to run the current tests...Any help would be appreciated :smile:

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -5,6 +5,7 @@ import ChildProcess from 'child_process'
 import resolveEnv from 'resolve-env'
 import { findCached } from 'atom-linter'
 import getPath from 'consistent-path'
+import untildify from 'untildify'
 
 const Cache = {
   ESLINT_LOCAL_PATH: Path.normalize(Path.join(__dirname, '..', 'node_modules', 'eslint')),
@@ -119,7 +120,14 @@ export function getArgv(type, config, filePath, fileDir, givenConfigPath) {
   if (config.eslintRulesDir) {
     let rulesDir = resolveEnv(config.eslintRulesDir)
     if (!Path.isAbsolute(rulesDir)) {
+      const splitPath = rulesDir.split('/')
+      if (splitPath[0] === '~') {
+        rulesDir = untildify(rulesDir)
+      }
       rulesDir = findCached(fileDir, rulesDir)
+    }
+    if (!rulesDir) {
+      throw new Error('Invalid ESLint Rules Dir path')
     }
     argv.push('--rulesdir', rulesDir)
   }
--- a/lib/worker-helpers.js
+++ b/lib/worker-helpers.js
@@ -30,6 +30,10 @@ var _consistentPath = require('consistent-path');
 
 var _consistentPath2 = _interopRequireDefault(_consistentPath);
 
+var _untildify = require('untildify');
+
+var _untildify2 = _interopRequireDefault(_untildify);
+
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
 
 var Cache = {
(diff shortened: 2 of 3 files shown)


Arcanemagus:
I know one of the other linters somewhat recently added this functionality... but I can't remember which one it was unfortunately.

kaicataldo:
Yeah, the script would definitely be helpful. Had no idea, but in retrospect that makes sense :D Hadn't realized apm can run tests. I can add that in as part of this PR if you want.

Arcanemagus:
Go for it if you want, had Travis rebuild btw and the current specs passed, it was failing due to a broken version of `eslint` that has since been moved on from.

kaicataldo:
Looked at a few and [untildify](https://www.npmjs.com/package/untildify looked the most promising). Thanks for the help :+1:

kaicataldo:
Huh, looks like a failed install failed the CI test

Arcanemagus:
Started a rebuild on there, not sure what happened.

Arcanemagus:
Any other comments @AtomLinter/linter-eslint?

kaicataldo:
Anything else I can do to land this? Looks like AppVeyor crashed while installing modules again, unfortunately :(

kaicataldo:
Rebased and everything is passing now :+1:

src/worker-helpers.js
```diff
@@ -119,7 +120,13 @@ export function getArgv(type, config, filePath, fileDir, givenConfigPath) {
   if (config.eslintRulesDir) {
     let rulesDir = resolveEnv(config.eslintRulesDir)
     if (!Path.isAbsolute(rulesDir)) {
-      rulesDir = findCached(fileDir, rulesDir)
+      const splitPath = rulesDir.split('/')
+      rulesDir = splitPath[0] === '~'
+        ? untildify(rulesDir)
```
IanVS: Does `untildify` perform any validations to check if the result actually exists?  If not, we may want to still pass the result to `findCached`.

kaicataldo: I'll look into this and report back

kaicataldo:
@IanVS Good call on checking if untildify actually checks for the existence of the file - updated!

kaicataldo:
Don't want to keep bugging you all, but just following up - is there anything else I can do to land this?

src/worker-helpers.js
```diff
@@ -119,8 +120,15 @@ export function getArgv(type, config, filePath, fileDir, givenConfigPath) {
   if (config.eslintRulesDir) {
     let rulesDir = resolveEnv(config.eslintRulesDir)
     if (!Path.isAbsolute(rulesDir)) {
+      const splitPath = rulesDir.split('/')
+      if (splitPath[0] === '~') {
```
Arcanemagus: Does `untildify` only work if it _begins_ with `~`? If not shouldn't we just pass it on if it contains `~` at all?

kaicataldo: Not sure what you mean - my understanding is that `~` is shorthand for the users home directory. A path like /Users/~ will try to find a directory called ~, which is a legitimate use case and shouldn't be expanded.  But maybe I'm misunderstanding what you're saying?

kaicataldo:
Don't want to keep bugging you all about this, but don't want to just leave it hanging either. Should I just close the PR?

kaicataldo:
Not sure if the implementation is wrong or if I'm missing something. Happy to reopen if this is a desired change - just let me know :)

PR #520, opened 2016-03-31
link react related rules to plugin docs

victorhaggqvist:
Links `react/` rules to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules instead of a 404 page.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -44,5 +44,12 @@ export function showError(givenMessage, givenDetail = null) {
 }
 
 export function ruleURI(ruleId) {
-  return `http://eslint.org/docs/rules/${ruleId}`
+  const ruleParts = ruleId.split('/')
+  if (ruleParts.length === 1) {
+    return `http://eslint.org/docs/rules/${ruleId}`
+  }
+  if (ruleParts[0] === 'react') {
+    return `https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/${ruleParts[1]}.md`
+  }
+  return ''
 }
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -61,5 +61,12 @@ function showError(givenMessage) {
 }
 
 function ruleURI(ruleId) {
-  return 'http://eslint.org/docs/rules/' + ruleId;
+  var ruleParts = ruleId.split('/');
+  if (ruleParts.length === 1) {
+    return 'http://eslint.org/docs/rules/' + ruleId;
+  }
+  if (ruleParts[0] === 'react') {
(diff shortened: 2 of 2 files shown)


src/helpers.js
```diff
@@ -44,5 +44,9 @@ export function showError(givenMessage, givenDetail = null) {
 }
 
 export function ruleURI(ruleId) {
+  if (ruleId.substring(0, 6) === 'react/') {
```
Arcanemagus: This might be clearer as `ruleId.startsWith('react/')`.

Arcanemagus:
Oooo, excellent idea!

Arcanemagus:
What about something like this?

```js
const ruleParts = ruleId.split('/')
if (ruleParts.length === 1) {
  return `http://eslint.org/docs/rules/${ruleId}`
}
if (ruleParts[0] === 'react') {
  return `https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/${ruleParts[1]}.md`
}
return ''
```

With that we just need to add a new if statement to add support for a new plugin's rules.

victorhaggqvist:
That seems like a good idea. I'll update the patch.

IanVS:
Seems like a good use of a `switch` statement, no?

victorhaggqvist:
@IanVS sound about right

victorhaggqvist:
Had another thought, it might be a good thing to keep plugins that have not been handled pointing to a 404, instead of do nothing. Since then people might realize it is something fixable (like I did). Thoughts?

(One could of course do a wikipage or something instead of 404, depending on the amount of effort worth putting into it..)

Arcanemagus:
That doesn't sound like a bad idea, a Wiki page on here detailing what would be necessary to add support for an unknown plugin.

Arcanemagus:
@victorhaggqvist Did you still want to put this together?

victorhaggqvist:
@Arcanemagus I would consider this PR complete. My comment (https://github.com/AtomLinter/linter-eslint/pull/520#issuecomment-203935222) was rather a suggestion for a possible future improvement.

IanVS:
@victorhaggqvist do you mind rebasing against master and fixing the conflicts, so we can merge?  Thanks!

@Arcanemagus I'll take a quick crack at a wiki, and a PR to point to it.

IanVS:
@victorhaggqvist are you interested in finishing this up?

victorhaggqvist:
@IanVS My bad, missed the notification. I'll look at it this weekend.

IanVS:
Thanks, @victorhaggqvist, but we've merged in https://github.com/AtomLinter/linter-eslint/pull/562 which included react plugin rules as well as several others.

PR #545, opened 2016-04-20
Fix: 'Fix on Save' should run only in valid file types

pvamshi:
We already have valid scopes listed as 

```js
this.scopes = ['source.js', 'source.jsx', 'source.js.jsx', 'source.babel', 'source.js-semantic']
```

I am using the same array and comparing if the file type belongs to one of them .
With this fix, it is not running in all file types. 
Please validate.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -93,7 +93,7 @@ module.exports = {
     }));
     this.subscriptions.add(atom.workspace.observeTextEditors(function (editor) {
       editor.onDidSave(function () {
-        if (atom.config.get('linter-eslint.fixOnSave')) {
+        if (_this.scopes.indexOf(editor.getGrammar().scopeName) !== -1 && atom.config.get('linter-eslint.fixOnSave')) {
           _this.worker.request('job', {
             type: 'fix',
             config: atom.config.get('linter-eslint'),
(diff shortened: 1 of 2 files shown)


IanVS:
It really seems like this should be more flexible, since we can't predict what file extensions everyone will use for javascript.  This is a good start to address the immediate bug, but I think a more flexible solution would be to lint (and fix) only `.js` files by default and provide an option for the user to configure others.

pvamshi:
Probably that should be valid for general linting. So that would directly change `this.scopes` array and this fix is still valid even in that case . 
Is my understanding correct ?

Arcanemagus:
This looks correct to me at least.

@IanVS I'm not sure what you are talking about here, this linter (and the changes in this PR) are based on the _scope_ as Atom sees it, not the extension on the file.

Or is there a limitation on `eslint`'s side that only pure JavaScript files can be fixed?

pvamshi:
Can we please merge this if the fix looks fine ?

IanVS:
Ah ok, I assumed the scope was determined by the extension.  Looks good to me then.

Arcanemagus:
That's essentially true for JavaScript, as the only place where I know of that `source.js` is valid is a pure `.js` file, but it's however Atom determines the scope really, which is controlled by the language packages.

Arcanemagus:
Published in v7.2.1. :tada:

PR #546, opened 2016-04-20
Move to eslint-config-airbnb-base

Arcanemagus:
Move to `eslint-config-airbnb-base` from `eslint-config-airbnb` as we don't need the React related stuff.

Updates the minimum Atom version to v1.4.0 so the `Object.assign()` polyfill can be removed. This could probably be earlier, but is there any reason to go further back?

Closes #540.
Closes #547.

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -1,6 +1,6 @@
-'use strict';
-// This file is used by eslint to hand the errors over to the worker
+"use strict";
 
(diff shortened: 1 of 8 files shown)


Arcanemagus:
@AtomLinter/linter-eslint This should probably get a review by somebody else before merging as it does change a few things.

IanVS:
Looks :+1: to me.

PR #562, opened 2016-05-01
Add URIs for documentation of plugins

relekang:
This makes the link to rule documentation work for the plugins used by the airbnb preset.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -61,5 +61,44 @@ function showError(givenMessage) {
 }
 
 function ruleURI(ruleId) {
-  return 'http://eslint.org/docs/rules/' + ruleId;
+  var ruleParts = ruleId.split('/');
+
+  if (ruleParts.length == 1) {
+    return 'http://eslint.org/docs/rules/' + ruleId;
+  }
+
+  switch (ruleParts[0]) {
+    case 'angular':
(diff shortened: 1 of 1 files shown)


relekang:
Should I add some more?

Arcanemagus:
Could you update the logic to similar to [here](https://github.com/AtomLinter/linter-eslint/pull/520/files#diff-27a1c86133de4310373decfdad200a91R64), but use a switch statement with a default case leading to [here](https://github.com/AtomLinter/linter-eslint/wiki/Linking-to-Rule-Documentation)?

Arcanemagus:
And please add as many as you want!

relekang:
Sure, just updated it :)

Arcanemagus:
@AtomLinter/linter-eslint Any further changes? This LGTM.

IanVS:
I love it.  Thanks, @relekang!

IanVS:
Shoot, I merged and then realized we're missing the `src` version.  The PR only had the compiled `lib` file, which will be overwritten as soon as someone compiles from source.  

Since it was my mistake, I'll make a commit to fix things up.

@Arcanemagus we should find a way to prevent this in the future.

relekang:
Oh, how did I miss this. Sorry about that and thank you for fixing it.

Arcanemagus:
Published in v7.2.2, thanks again @relekang!

PR #566, opened 2016-05-03
Add plugin rule switch statement to src

IanVS:
I merged #562 too quickly before realizing that there were no `src` file
changes, only changes to `lib`, which would have been overwritten the
first time someone else compiled `src` to `lib`.  This commit brings
@relekang’s contribution over to the `src` file with a few minor
modifications to meet our linting rules.


IanVS:
@Arcanemagus I plan to merge as soon as tests pass.

Arcanemagus:
Whoops, should have caught that myself!

Issue #548, opened 2016-04-21
disableEslintIgnore doesn't seem to be respected


bojand:
What is exactly `Disable using .eslintignore files` checkbox / setting suppose to to? Disable the plugin just if I have `.eslintignore` file present? If so it doesn't seem to be working. My config:

```
"linter-eslint":
    disableEslintIgnore: true
    disableWhenNoEslintConfig: false
    useGlobalEslint: true
```

I have `.eslintignore` present in the root of my project, yet I am still getting an error:

![image](https://cloud.githubusercontent.com/assets/2081100/14712256/c290e55c-07b2-11e6-8216-72ff46b07b53.png)

I've tried having an empty `.eslintignore` or one that has just a valid random folder, ie. `node_modules`.
Am I doing something wrong?

Thanks,

IanVS:
If you check that setting (`disableEslintIgnore`), we won't look at the `.eslintignore` file at all.  Can you explain what you're trying to do?

bojand:
Hello, thanks. I must have a misunderstanding what that setting / checkbox is for and it's suppose to do at all? My assumption was that if you have `.eslintignore` present, it would turn off `atom-eslinter` plugin? Is this correct? What is the meaning of this setting?

What I am trying to do is the following...

In some of my projects I use [xo](https://github.com/sindresorhus/xo) and its [Atom plugin](https://github.com/sindresorhus/atom-linter-xo). I have `eslint` and `xo` globally installed. `xo` uses `eslint`. In the cases where I have `xo` setup I would like linting through `atom-eslinter` package to be turned off _somehow_. So I had assumed that just putting `.eslintignore` would turn off the plugin?

In cases where I do not have `xo` setup in the project, I would like linting using (my global) `eslint` and `atom-eslinter` plugin to be done, regardless if I have `.eslintrc` or ESLint config in `package.json`.

It seems I am assuming what that setting / checkbox _actually means_?

Thanks,

IanVS:
An `.eslintignore` file is used by ESLint to disable linting for particular files.  For more information on how they work, check out http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories

Do the projects which use xo have `.eslintrc` files in them?  It seems like you wouldn't since the top of the xo readme says:

> No decision-making. No `.eslintrc` or `.jshintrc` to manage. It just works!

If that's the case, you can use `disableWhenNoEslintConfig` to accomplish what you want.  

[...]

PR #569, opened 2016-05-06
Clarified disableEslintIgnore description

Zacqary:
As brought up in https://github.com/AtomLinter/linter-eslint/issues/548, `Disable using .eslintignore files` is very ambiguous. It's not clear whether checking the box will "Disable (the use of) .eslintignore files," (which is actually what it's doing) or whether checking the box will "Disable (this package) using .eslintignore files" (which makes it appear to be a bug when this box isn't checked but the package still obeys `.eslintignore` files).

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -54,7 +54,7 @@ module.exports = {
       default: ''
     },
     disableEslintIgnore: {
-      title: 'Disable using .eslintignore files',
+      title: 'Don\'t use .eslintignore files',
       type: 'boolean',
       default: false
     },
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Just because it amuses me, what do you think about `Ignore .eslintignore files`? :stuck_out_tongue:

Arcanemagus:
Just to be clear, I'm fine with either wording, I just came up with that when looking over the changes.

Zacqary:
I think they're both less confusing, but the fact that `Ignore .eslintignore files` is hilarious might be too distracting.

Arcanemagus:
Works for me merging and thanks!

PR #609, opened 2016-06-14
Allow user to specify a local node_modules path. Continues #585

MikeRatcliffe:
Sorry, had a whole host of issues with my git database so I have had to create a new pull request.

I have now followed jsnajdr's config comments from (here)[https://github.com/AtomLinter/linter-eslint/pull/585#issuecomment-221630361].

Both absolute and relative paths are valid.


src/main.js
```diff
@@ -40,6 +40,12 @@ module.exports = {
       type: 'string',
       default: ''
     },
+    localNodeModules: {
+      title: 'Path to the local node_modules folder',
+      description: 'Optionally specify the path to the local node_modules folder',
```
Arcanemagus: Maybe put `, can be relative to the current project` on the end so users know about that?

Arcanemagus:
Are the `.gitkeep` files necessary in folders that have sub-folders with "real" files in them? (Like `spec/fixtures/indirect-local-eslint/testing/eslint/node_modules/eslint/.gitkeep`)

Arcanemagus:
Other than the above, this LGTM. @AtomLinter/linter-eslint any other comments?

IanVS:
No comments from me other than to [reiterate](https://github.com/AtomLinter/linter-eslint/pull/585#issuecomment-221568452) that I think we need some explanatory documentation around this option.

MikeRatcliffe:
@IanVS Which documentation would you like updated? The documentation under https://github.com/AtomLinter/linter-eslint/wiki?

Arcanemagus:
> I'd like to see an update to the README explaining what it is and the use cases it's intended to address.

I'm assuming this is what he means, just add a "Configuration" section and describe what the option does. We try to keep people using the Settings UI, so don't provide it in a form that can be directly copy/pasted into a `config.cson` or people will do _exactly_ that, and then complain when that doesn't work properly 😛.

IanVS:
I was thinking more about adding a section to the `README.md` to explain the options and what they mean/do.

past:
What is blocking this from merging? Are there any issues still unresolved? It's frustrating to work on Firefox frontend code with Atom, without this fix.

Arcanemagus:
@past A combination of commits alone not triggering notifications back when this was last updated, and a lack of attention to the PR on our part, for which I'm sorry!

We're making a push to get things in for a major release and it would be nice to be able to get this in. @MikeRatcliffe would you be interested in rebasing this on the current master? I think this PR is too old for the ability to be enabled for us to commit to this branch unfortunately.

MikeRatcliffe:
@past: Just been busy with browser.html
@Arcanemagus: Of course, I will rebase it as soon as I get time.

Arcanemagus:
@MikeRatcliffe I'm having a little trouble understanding what you did here... it looks like you cherry picked the two commits, and then for some reason merged in the original two commits... along with all the commits to `master`?

Is there any reason you don't just reset the branch to where you have dfa2f7d? That looks like a good point to start reviewing to me.

MikeRatcliffe:
>I'm having a little trouble understanding what you did here... it looks like you cherry picked the two commits, and then for some reason merged in the original two commits... along with all the commits to master?

Yes, the repo turned out in a mess and then I tried to fix the travis-ci error (eslint-plugin-import version) until I realized that we are locked into v.1.16.0 because it is a dependancy of eslint-config-airbnb-base.

>Is there any reason you don't just reset the branch to where you have dfa2f7d? That looks like a good point to start reviewing to me.

@Arcanemagus: No good reason... done!

spec/linter-eslint-spec.js
```diff
@@ -4,6 +4,7 @@ import * as path from 'path'
 import * as fs from 'fs'
 import { tmpdir } from 'os'
 import rimraf from 'rimraf'
+// eslint-disable-next-line import/no-named-as-default, import/no-named-as-default-member
```
Arcanemagus: I'm guessing this line snuck in from the `eslint-plugin-import@2.0.0` part, it's not necessary (yet), so it should be removed.

spec/worker-helpers-spec.js
```diff
@@ -6,6 +6,27 @@ import { getFixturesPath } from './common'
 
 describe('Worker Helpers', () => {
   describe('getESLintInstance && getESLintFromDirectory', () => {
+    it('tries to find an indirect local eslint using an absolute path', () => {
+      const path = Path.join(getFixturesPath('indirect-local-eslint'),
```
Arcanemagus: I'd prefer the split to be right after `Path.join(` myself.

spec/worker-helpers-spec.js
```diff
@@ -6,6 +6,27 @@ import { getFixturesPath } from './common'
...
+                       'testing', 'eslint', 'node_modules')
+      const eslint = Helpers.getESLintInstance('', {
+        useGlobalEslint: false,
+        localNodeModules: path
+      })
+      expect(eslint).toBe('located')
+    })
+    it('tries to find an indirect local eslint using a relative path', () => {
+      const path = Path.join(getFixturesPath('indirect-local-eslint'),
```
Arcanemagus: Same here.

src/main.js
```diff
@@ -45,10 +45,14 @@ module.exports = {
       editor.onDidSave(() => {
         if (scopes.indexOf(editor.getGrammar().scopeName) !== -1 &&
             atom.config.get('linter-eslint.fixOnSave')) {
+          const filePath = editor.getPath()
+          const projectPath = atom.project.relativizePath(filePath)[0] || ''
```
Arcanemagus: Take off the ` || ''` here, this is already being done in `worker-helpers.js`.

src/main.js
```diff
@@ -60,6 +64,7 @@ module.exports = {
       'linter-eslint:fix-file': () => {
         const textEditor = atom.workspace.getActiveTextEditor()
         const filePath = textEditor.getPath()
+        const projectPath = atom.project.relativizePath(filePath)[0] || ''
```
Arcanemagus: Also take off the ` || ''` here.

src/main.js
```diff
@@ -121,18 +127,21 @@ module.exports = {
           return Promise.resolve([])
         }
         const filePath = textEditor.getPath()
+        const projectPath = atom.project.relativizePath(filePath)[0] || ''
+        const config = atom.config.get('linter-eslint')
```
Arcanemagus: As this isn't being used elsewhere I don't really see a reason to define it as an explicit variable. (Passing things properly is a longer term goal of mine...)

package.json
```diff
@@ -43,6 +43,12 @@
       "type": "string",
       "default": ""
     },
+    "localNodeModules": {
+      "title": "Path to the local node_modules folder",
+      "description": "Optionally specify the path to the local node_modules folder",
```
Arcanemagus: Maybe prefix this with `Advanced: `? Something to indicate this is an uncommon feature and is likely something the user doesn't need except in very specific circumstances.

spec/linter-eslint-spec.js
MikeRatcliffe: Removed

spec/worker-helpers-spec.js
MikeRatcliffe: Fixed

spec/worker-helpers-spec.js
MikeRatcliffe: Fixed

src/main.js
MikeRatcliffe: Done

src/main.js
MikeRatcliffe: Done

src/main.js
MikeRatcliffe: Fixed

MikeRatcliffe:
@Arcanemagus I have addressed all of your review comments so I am guessing that this should be ready to land.

Arcanemagus:
Merged, thanks!

PR #629, opened 2016-07-02
Make scopes a configuration setting

neurosnap:
This PR will allow the user to add or remove scopes that will apply to `linter-eslint`.  The primary reason I am submitting this PR is for https://github.com/rtsao/csjs and by proxy https://github.com/neurosnap/language-csjs which will allow us to extend the linter to this pseudo language.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,14 @@
       "description": "Have eslint attempt to fix some errors automatically when saving the file.",
       "type": "boolean",
       "default": false
+    },
+    "scopes": {
+      "title": "Language scopes to use linter",
+      "type": "array",
+      "default": ["source.js", "source.jsx", "source.js.jsx", "source.babel", "source.js-semantic"],
+      "items": {
+        "type": "string"
+      }
     }
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
       "default": false
     },
     "scopes": {
-      "title": "Language scopes to use linter",
+      "title": "List of scopes to run ESLint on, run `Editor: Log Cursor Scope` to determine the scopes for a file.",
       "type": "array",
       "default": [
         "source.js",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,13 @@
     "scopes": {
       "title": "Language scopes to use linter",
       "type": "array",
-      "default": ["source.js", "source.jsx", "source.js.jsx", "source.babel", "source.js-semantic"],
+      "default": [
+        "source.js",
+        "source.jsx",
+        "source.js.jsx",
+        "source.babel",
+        "source.js-semantic"
+      ],
       "items": {
         "type": "string"
       }
--- a/src/main.js
+++ b/src/main.js
@@ -11,8 +11,14 @@ module.exports = {
     this.subscriptions = new CompositeDisposable()
     this.active = true
     this.worker = null
-    this.scopes = atom.config.get('linter-eslint.scopes')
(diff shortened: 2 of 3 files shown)


src/main.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     this.subscriptions = new CompositeDisposable()
     this.active = true
     this.worker = null
-    this.scopes = ['source.js', 'source.jsx', 'source.js.jsx', 'source.babel', 'source.js-semantic']
+    this.scopes = atom.config.get('linter-eslint.scopes')
```
Arcanemagus: Maintaining the scopes list isn't _quite_ this simple. Your current code will only update on launching Atom, since that is (generally) the only time the `activate()` function is called.

You'll need to update this to code like [this](https://github.com/AtomLinter/linter-proselint/blob/483fa1d7c65f7f83a1995a7aa4b546a734378bab/lib/index.js#L12-L17), so whenever the setting is changed, the scopes array is update (_not_replaced!) with the new values.

package.json
```diff
@@ -65,6 +65,14 @@
       "description": "Have eslint attempt to fix some errors automatically when saving the file.",
       "type": "boolean",
       "default": false
+    },
+    "scopes": {
+      "title": "Language scopes to use linter",
+      "type": "array",
+      "default": ["source.js", "source.jsx", "source.js.jsx", "source.babel", "source.js-semantic"],
```
Arcanemagus: Separate these on to individual lines (`npm install --save <package>" will re-write the file like this, meaning the next GreenKeeper PR will do so, might as well get it right from the start).

package.json
```diff
@@ -65,6 +65,14 @@
       "description": "Have eslint attempt to fix some errors automatically when saving the file.",
       "type": "boolean",
       "default": false
+    },
+    "scopes": {
+      "title": "Language scopes to use linter",
+      "type": "array",
```
Arcanemagus: Add a description along the lines of [this](https://github.com/AtomLinter/linter-proselint/blob/483fa1d7c65f7f83a1995a7aa4b546a734378bab/package.json#L15) describing what this setting does, and how the user can get the information to update it.

package.json
```diff
@@ -65,6 +65,14 @@
       "description": "Have eslint attempt to fix some errors automatically when saving the file.",
       "type": "boolean",
       "default": false
+    },
+    "scopes": {
+      "title": "Language scopes to use linter",
```
Arcanemagus: Change "use linter" to something like "run ESLint on".

src/main.js
```diff
@@ -11,8 +11,14 @@ module.exports = {
     this.subscriptions = new CompositeDisposable()
     this.active = true
     this.worker = null
-    this.scopes = atom.config.get('linter-eslint.scopes')
+    this.scopes = new CompositeDisposable()
```
Arcanemagus: `this.scopes` _must_ be an `Array`

src/main.js
```diff
@@ -11,8 +11,14 @@ module.exports = {
...
 
+    this.scopes.add(atom.config.observe('linter-eslint.scopes', scopes => {
```
Arcanemagus: You should be adding this `Disposable` to `this.subscriptions`.

Arcanemagus:
@neurosnap A few more changes needed here.

Arcanemagus:
LGTM and CI is green, merging 😉.

Thanks for putting this together!

neurosnap:
I'm glad we could get this merged.  Thanks for all the input and examples, much appreciated!

PR #653, opened 2016-07-29
Add eslint-plugin-standard rule docs link

apexskier:
This package is currently missing docs links for [eslint-plugin-standard](https://github.com/xjamundx/eslint-plugin-standard#rules-explanations) so I'm adding them! The specific rule I hit was `standard/array-bracket-even-spacing`.

(I got here from https://github.com/AtomLinter/linter-eslint/wiki/Linking-to-Rule-Documentation#adding-rule-doc-links.)

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -83,6 +83,9 @@ export function ruleURI(ruleId) {
     case 'react':
       return `https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/${ruleName}.md`
 
+    case 'standard':
+      return 'https://github.com/xjamundx/eslint-plugin-standard#rules-explanations'
+
     default:
       return 'https://github.com/AtomLinter/linter-eslint/wiki/Linking-to-Rule-Documentation'
   }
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -83,6 +83,9 @@ export function ruleURI(ruleId) {
     case 'react':
       return `https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/${ruleName}.md`
 
+    case 'standard':
+      return `https://github.com/xjamundx/eslint-plugin-standard#rules-explanations`
```
IanVS: You can use single-quotes here, backticks are only allowed by our linting config when using features like string interpolation (e.g. on the line above).

IanVS:
Thanks!  I guess I should finish up that wiki...  Glad you figured it out even though it's not done.

Also, it's annoying, but we also need to check in the compiled `lib` directory.  Would you mind running `npm run compile` and committing the corresponding changes in `lib` as well?

apexskier:
Sure, I just did this in GitHub's editor. I won't be able to update until Monday after next, but I'll run through the linter and compile and push up then.

apexskier:
Updated and tests passing!

Arcanemagus:
Can you make this change over in [eslint-rule-documentation](https://github.com/jfmengels/eslint-rule-documentation)? As soon as #657 lands all that code will be moved there 😉.

Sorry for not commenting that sooner!

apexskier:
Ha, sure no problem

apexskier:
https://github.com/jfmengels/eslint-rule-documentation/pull/2

PR #656, opened 2016-08-01
make config clearer in README.md

braco:


README.md
```diff
@@ -38,8 +38,16 @@ You have two options:
 *   Install globally `eslint` and plugins
 
     *   `$ npm i -g eslint [eslint-plugins]`
-    *   Activate `Use Global Eslint` package option
+    *   Activate `Use Global Eslint` package option in Sublime's config
```
Arcanemagus: This isn't Sublime 😛.

braco: @Arcanemagus lol! oops.

README.md
```diff
@@ -38,8 +38,16 @@ You have two options:
...
+    ```
```
Arcanemagus: Exact code like this is _specifically_ not in there, because you should just be using the Settings UI to change it in the first place.

For anyone who feels the need to edit that file manually, they should intimately know the format and how to find the setting values themselves.

README.md
```diff
@@ -38,8 +38,16 @@ You have two options:
...
+      "linter-eslint":
+        "useGlobalEslint": true
+    ```
     *   (Optional) Set `Global Node Path` with `$ npm config get prefix`
+    ```
+      "linter-eslint":
+        "globalNodePath": "/usr/local"
```
Arcanemagus: This is a _perfect_ example as to how putting copyable examples in the readme is a bad idea. This may be valid for your system, but other *NIX systems may place it in a different folder, and that isn't even a valid path on Windows systems.

PR #665, opened 2016-08-12
Add config option to specify a custom eslint installation directory

jstejada:
This commit adds a new config option to specify a custom eslint
installation directory, which can be a path relative to the atom
project, or a global path.

If the option is provided, it takes precendence over the global option,
and will be used as a first option instead of a global or local installations.

Adds specs and fixes a test that was incorrectly passing


spec/worker-helpers-spec.js
```diff
@@ -7,42 +7,58 @@ import * as Path from 'path'
...
     it('tries to find a local eslint', () => {
-      const eslint = Helpers.getESLintInstance(getFixturesPath('local-eslint'), {})
+      const fileDir = getFixturesPath('local-eslint')
+      const eslint = Helpers.getESLintInstance(fileDir, '', {})
       expect(eslint).toBe('located')
     })
-    it('cries if local eslint is not found', () => {
-      expect(() => {
-        Helpers.getESLintInstance(getFixturesPath('files', {}))
-      }).toThrow()
```
jstejada: This was actually throwing the wrong error: `Can't read property useGlobalEslint of undefined` because the `{}` was not being correctly passed in to `getESLintInstance`.

I decided to remove this test completely because it will always find our own installation of eslint (the one in `linter-eslint/node_modules`).

jstejada:
@steelbrain do you have any comments on this? This option would be really helpful for my project setup.

steelbrain:
@jstejada while it's implemented nicely, I think that we should be Node.js behavior compliant.

Here's how it works ( or at least should work )

1. Checks for any local ESLint installations, uses them
2. Falls back to bundled ESLint

What you are proposing is that

1. Use the config defined ESLint installation
2. Fallback to local
3. Fallback to bundled

I wouldn't have minded it if configs were per-project but configs are globals so by fixing this for one project, you're breaking it for all the rest. Furthermore ESLint urges you to install it locally (or link the correct verison locally at least) so you don't run into issues, if you follow their recommendation, following the current pattern of execution you shouldn't have any issues

Arcanemagus:
As far as I can tell, this is accomplishing the same thing as #609, as such I'm going to close this in favor of that PR.

If you think that #609 won't satisfy the reason you created this PR @jstejada can you please elaborate on what you are trying to accomplish?

jstejada:
sounds good, thanks @Arcanemagus

Arcanemagus:
Thanks for putting the effort into this, and I'm sorry it didn't get any attention earlier!

Issue #621, opened 2016-06-26
Allow rules overriding


ngryman:
Hi,

In my use case I report the `no-multiple-spaces` as an error. But for obvious reasons, I don't want this rule to be reported while I'm coding. Still, I want to keep the default cascading behavior and my project `eslintrc`.

Being able to specify a config path that would take precedence on any other rules would be perfect. That would allow to tweak the config for a live coding environment.

Thanks!

![eslint](https://cloud.githubusercontent.com/assets/892048/16360004/87ba7648-3b4a-11e6-97ab-338d43f10c1b.gif)

Arcanemagus:
Is this even possible @IanVS? As far as I know `eslint` doesn't support anything like this.

ngryman:
@Arcanemagus I suppose there are several ways of forcing a set of rules.

One would be to use the `--rule` command line switch of `eslint` as it has a higher precedence (http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy).
But it would be cumbersome I guess.

An other would be to extend the found project configuration with the global one and pass the result to `eslint`.

There are probably other ways of achieving this but I'm not aware of them.

---

On a side note, I think this should be an important feature. Linting a *live* code and a codebase should be treated differently as the context is different.
*Live* code evolves and while you are typing you are certainly violating some rules that only make sense when the code is in its final version.

Another *simpler* way to solve this problem, would be to allow the user to lint only on save, in order to review the code when he wants.

Arcanemagus:
Interesting idea, I can see allowing an array of `--rule` specifications to pass in only for on the fly lints, while letting it use the unaltered project configuration for a "on save" lint, that would require an almost complete re-write of the linter though as it would need to move to the Indie API to be able to provide different functionality based on the type of lint call...

sradevski:
What I was thinking of is quite simple: Have an option that will only lint on save. Like this you can code without the annoying linting messages, and once saved, the whole file will be linted and it will show the linting errors.

ngryman:
@sradevski This option already exist in the `linter` package as `Lint As You Type`.

sradevski:
@ngryman I wasn't aware of it, thanks!

IanVS:
@ngryman would https://github.com/AtomLinter/linter-eslint/pull/666 meet your needs?

ngryman:
@IanVS It seems perfect! I guess we can close this one then.

jakedeichert:
[...]

Issue #596, opened 2016-06-07
Disable verbose error display for  trailing whitespace rule violation on current line


ddumontatibm:
![image](https://cloud.githubusercontent.com/assets/648319/15873969/062dd298-2ccf-11e6-80dd-4dc2b9c8e0ce.png)

I like the linter... but it kinda gets in the way in this case.   I don't mind the red lines... but the popup should perhaps be suppressed in this specific case.

Arcanemagus:
We don't actually have anything to do with the UI, this package just provides a translation between ESLint and Linter. You might want to increase the delay in the Linter settings so it doesn't lint right away, I personally use 1500 ms for the delay.

Arcanemagus:
You should follow #666, as it will allow you to hide that rule when editing the file (but still show it for on-save lints).

PR #666, opened 2016-08-12
Add setting for silencing rule ids

einarlove:
This PR adds the support for silencing/turn off rules for linter-eslint. Useful when some rules are important but annoying while writing, like `no-trailing-spaces` or `eol-last` since they will be fixed on save by default Atom settings.


IanVS:
Hm, interesting idea.  I haven't given the code itself a review yet, but I can see the use case for having a set of rules that you don't want to be warned about while you're typing.  

What do you think, @Arcanemagus?

Arcanemagus:
See https://github.com/AtomLinter/linter-eslint/issues/621#issuecomment-228853104 for my thoughts 😛.

tl;dr is that we would need to move to the Linter Indie API to be able to provide different results for the on-the-fly linting vs the on-save linting. There is one other linter that currently works this way, for exactly this reason, it's a complex task though.

einarlove:
Having different rules silenced in different contexts complicates this simple feature, but is that what you would want? I could investigate and see if we can limit it to only when you type?

Arcanemagus:
At least with the way I use this I want `linter-eslint` to be telling me about anything that would be shown if I ran `eslint` on the CLI. The way I was envisioning this working in the other issue is you could hide messages for lint-on-fly lints, but they would still show when the file is saved. This would prevent any "gotchas" later on when you try pushing up a PR for example.

Maybe I'm not the target for this functionality though, so if @IanVS is okay with this being a blanket ignore, with the possibility of things showing up later outside of the Atom environment, I guess I would be fine with merging just that.

einarlove:
It now only silence rules while typing!

Arcanemagus:
Wow, well that was much simpler than I was thinking 😁.

package.json
```diff
@@ -79,6 +79,15 @@
       "items": {
         "type": "string"
       }
+    },
+    "silenceWhileTypingRules": {
+      "title": "Silence rules while typing",
```
Arcanemagus: Maybe "Rules to silence while typing"? The current wording implies _all_ rules are silenced while typing, and that this should be a flag.

package.json
```diff
@@ -79,6 +79,15 @@
...
+      "description": "Useful when Atom fixes errors on save like no-trailing-spaces or eol-last.",
```
Arcanemagus: Wrap the rule names in \`.

src/main.js
```diff
@@ -98,11 +98,17 @@ module.exports = {
         }
         const filePath = textEditor.getPath()
         const showRule = atom.config.get('linter-eslint.showRuleIdInMessage')
+        let rules = {}
+
+        if (textEditor.isModified()) {
+          rules = idsToIgnoredRules(atom.config.get('linter-eslint.silenceWhileTypingRules'))
```
Arcanemagus: Observe the setting in `activate` so it only has to do the complex parsing when things change.

src/helpers.js
```diff
@@ -43,6 +43,13 @@ export function showError(givenMessage, givenDetail = null) {
   })
 }
 
+export function idsToIgnoredRules(ruleIds = []) {
+  return ruleIds.reduce((ids, id) => {
+    ids[id] = 0
```
IanVS: I think it's good to use `0` like you have, to support older versions of ESLint which didn't understand `off`, but maybe it would be better to make a const like `const RULE_OFF_SEVERITY = 0` to avoid the magic number?

src/worker-helpers.js
```diff
@@ -126,6 +126,9 @@ export function getArgv(type, config, filePath, fileDir, givenConfigPath) {
   if (configPath) {
     argv.push('--config', resolveEnv(configPath))
   }
+  if (rules) {
+    argv.push('--rule', JSON.stringify(rules))
```
Arcanemagus: Doesn't this option specify the rules to _include_, not exclude?

src/main.js
```diff
@@ -98,11 +98,17 @@ module.exports = {
...
+        }
 
         return this.worker.request('job', {
           contents: text,
           type: 'lint',
           config: atom.config.get('linter-eslint'),
+          rules,
```
Arcanemagus: Move this below `filePath` to match the order where it's used in `worker.js`.

src/main.js
IanVS: Maybe it would be more clear to call this variable `ignoredRules`?  I imagine this might be confusing for others approaching the code without knowing the background of why it's being done.

Arcanemagus: The rules aren't _always_ ignored though, so the name should include some indication that it is only for modified files.

IanVS:
Awesome work!  I had a few minor suggestions, but I'm excited to see this land.

Arcanemagus:
Looks like this also needs some specs 😉.

einarlove:
New to specs and testing so might take a while before i figure out how it works. Will look into it this afternoon.

einarlove:
Well this atom test setup was horrible. Incredible frustrating when the errors messages in `apm test` are so cryptic. Would love if someone could help with writing the spec to this, if it's needed.

Arcanemagus:
It's much easier to run them within Atom (View -> Developer -> Run Package Specs) since you can then open developer tools on the spec window and get full debugger access. Note that the spec window uses the default keybinds if you have customized the one to open the developer tools.

IanVS:
@einarlove are you still stuck on spec tests for this?  Let's see what we can do to help get this merged.  Are there any areas in particular that you're having trouble with?

einarlove:
Yeah my motivation plummeted while writing the specs. Give a try but could not make the specs I wrote work.

Would love if someone helped write the specs.

Arcanemagus:
@einarlove If you enable the new ability for contributors to the base repo to be able to commit to PR branches on forks (check mark on the bottom of the right side bar) one of us can take a stab at it and commit directly here 😉.

ddumontatibm:
From #596, please note that while Atom will fix line endings on save, it will not remove trailing whitespace from a line if the cursor is *on* it and the file is saved. (see picture in #596)

So this doesn't completely solve the problem, but I guess it does make it a bit easier to live with.

einarlove:
@Arcanemagus Done! 👍🏼

Arcanemagus:
Hmmm, this is passing locally, time to look further a bit...

Arcanemagus:
Something is very flaky with this spec, now CircleCI is passing, but _half_ of Travis is passing, half isn't: https://travis-ci.org/AtomLinter/linter-eslint/builds/161776266

src/main.js
```diff
@@ -70,6 +70,10 @@ module.exports = {
       }
     }))
 
+    this.subscriptions.add(atom.config.observe('linter-eslint.rulesToSilenceWhileTyping', (ids) => {
+      this.ignoredRulesWhenModified = idsToIgnoredRules(ids)
+    }))
```
IanVS: I'm trying to get the spec working, and finding it very difficult.  The core of the problem is that this subscription is what determines which rules are ignored when the file is modified.  In our current specs, this subscription is never added. When I try to call `linter.activate()` so the subscription is added, I get an error from the first line of the method which calls `require('atom-package-deps').install()`.  If I comment that out, the specs seem to work fine except that changing the config to add a rule to `rulesToSilenceWhileTyping` still is not observed (no idea why not) and therefore `this.ignoredRulesWhenModified` is never changed.  

So, long-and-short, I'm a bit stumped on how to proceed here.  I will push up a change that at least gets this rebased on master and makes the spec run correctly, but it will be failing.  Maybe someone else can take it from there.

einarlove:
No wonder I got burned out trying to make these specs to pass 😅

Arcanemagus:
Yea, turns out things were in far worse need of attention than I thought... _hopefully_ this last set of fixes gets things working again.

This PR has ballooned way out of scope, but I'm not sure I care enough right now to split it up like it should be. We can do that if you feel we should though @IanVS.

Arcanemagus:
@einarlove We're splitting out the bugfixing into separate PR's, I'll be rebasing this and cleaning it up once those get merged.

Arcanemagus:
Looks like the expected value fixes got lost in the rebase, whoops!

Arcanemagus:
This PR is (finally) ready for merge, pending CI and @IanVS approval 😁.

PR #657, opened 2016-08-01
Use `eslint-rule-documentation` to get rule documentation links

jfmengels:
Use `eslint-rule-documentation` to get rule documentation links, as proposed in #590.

This was pretty straightforward, but this part didn't have tests and I'm not sure how to test this in Atom. Can I ask you for guidance in order to test this properly? (Unless you find this too straightforward and you can test it easily)

PR #675, opened 2016-08-20
add react-native doc link

mcecot:
i could not manage to test this pr. either some magic happens here or i should first issue missing documentation on that. could you explain how to test or to implement a further doc link? the according wiki page ends abruptly.


Arcanemagus:
Can you file this over on [`eslint-rule-documentation`](https://github.com/jfmengels/eslint-rule-documentation)? As soon as #657 gets merged that code will no longer live here.

Arcanemagus:
Actually looks like it's already over there 😛.

PR #694, opened 2016-09-10
throw original linter error when library throws range error

dmueller39:


IanVS:
Works great, but unfortunately our build process is a bit annoying and we have to check in our compiled files, which is what you have edited (which would then be overwritten on the next compile).

Could you please instead make those same changes to `src/main.js`, and then run `npm compile` and commit the changes to both files?

neurosnap:
I don't mean to hijack your PR @dmueller39 but getting this merged will help us track down the errors we are seeing in the linter, so I made the fixes that @IanVS suggested in #696

IanVS:
Looks good, thanks!

PR #696, opened 2016-09-12
Throw original error when range error is thrown

neurosnap:
Fixes issues with PR #694


dmueller39:
@neurosnap just updated mine, but it looks like yours might be more in line with the projects conventions.

neurosnap:
@dmueller39 im going to close my issue, yours looks good, you just need to run `npm compile`

PR #709, opened 2016-09-21
Add support for endLine and endColumn

Arcanemagus:
In https://github.com/eslint/eslint/pull/6640, released in `eslint@3.1.0`, ESLint added the ability for a rule to return a full range instead of just a single point. This PR adds support for that.

Merging of this is currently blocked however as no core rule currently actually returns a range (https://github.com/eslint/eslint/issues/3307), so we have nothing to validate against.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -7,6 +7,7 @@ Object.defineProperty(exports, "__esModule", {
 exports.spawnWorker = spawnWorker;
 exports.showError = showError;
 exports.idsToIgnoredRules = idsToIgnoredRules;
+exports.validatePoint = validatePoint;
 
 var _child_process = require('child_process');
 
@@ -73,4 +74,12 @@ function idsToIgnoredRules() {
     ids[id] = RULE_OFF_SEVERITY;
     return ids;
   }, {});
+}
+
+function validatePoint(textEditor, line, col) {
+  var buffer = textEditor.getBuffer();
+  // Clip the given point to a valid one, and check if it equals the original
+  if (!buffer.clipPosition([line, col]).isEqual([line, col])) {
+    throw new Error(line + ':' + col + ' isn\'t a valid point!');
+  }
 }
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -50,7 +50,7 @@ function spawnWorker() {
 }
 
 function showError(givenMessage) {
-  var givenDetail = arguments.length <= 1 || arguments[1] === undefined ? null : arguments[1];
+  var givenDetail = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : null;
 
   var detail = void 0;
   var message = void 0;
@@ -68,7 +68,7 @@ function showError(givenMessage) {
 }
 
 function idsToIgnoredRules() {
-  var ruleIds = arguments.length <= 0 || arguments[0] === undefined ? [] : arguments[0];
+  var ruleIds = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : [];
 
   return ruleIds.reduce(function (ids, id) {
     ids[id] = RULE_OFF_SEVERITY;
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -56,3 +56,11 @@ export function idsToIgnoredRules(ruleIds = []) {
     return ids
   }, {})
 }
+
+export function validatePoint(textEditor, line, col) {
+  const buffer = textEditor.getBuffer()
+  // Clip the given point to a valid one, and check if it equals the original
+  if (!buffer.clipPosition([line, col]).isEqual([line, col])) {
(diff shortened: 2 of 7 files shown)


Arcanemagus:
Note: This is currently based on the branch for #692 and should be rebased on `master` once that gets merged.

lib/helpers.js
```diff
@@ -61,3 +62,11 @@ function showError(givenMessage) {
     dismissable: true
   });
 }
+
+function validatePoint(textEditor, line, col) {
```
Arcanemagus: Should this be `validateRange` and do similar logic but just use `TextBuffer::clipRange`? It would save calling it twice...

Arcanemagus:
Looks like `no-unreachable` supports this format, writing a spec now.

src/main.js
```diff
@@ -164,10 +166,16 @@ module.exports = {
               }
             }
             let range
+            const msgLine = line - 1
+            const msgCol = column ? column - 1 : column
```
IanVS: This works fine, but it might be nice to be a little more explicit with something like:

```js
const msgCol = column > 0 ? column - 1 : 0
```

or even better (I think):

```js
const msgCol = Math.max(0, column - 1)
```

src/main.js
```diff
@@ -164,10 +166,16 @@ module.exports = {
...
             try {
-              range = Helpers.rangeFromLineNumber(
-                textEditor, line - 1, column ? column - 1 : column
-              )
+              if (endColumn && endLine) {
+                validatePoint(textEditor, msgLine, msgCol)
+                validatePoint(textEditor, endLine - 1, endColumn - 1)
+                range = [[msgLine, msgCol], [endLine - 1, endColumn - 1]]
```
IanVS: Question: why are you subtracting 1 from the `endColumn`?  I understand subtracting for `endLine`, since it is 1-based, but `endColumn` is 0-based.

src/main.js
```diff
@@ -164,10 +166,16 @@ module.exports = {
               }
             }
             let range
+            const msgLine = line - 1
+            const msgCol = column ? column - 1 : column
             try {
-              range = Helpers.rangeFromLineNumber(
-                textEditor, line - 1, column ? column - 1 : column
-              )
+              if (endColumn && endLine) {
```
IanVS: Are we sure that a valid `endColumn` won't be `0`?  If we get an endColumn of zero, and an `endLine` of, say, `5`, it seems like we should still use them.  But I think this check will return false, right?  Perhaps a safer thing to do is check `typeof endColumn !== 'undefined'`?

src/helpers.js
```diff
@@ -56,3 +56,11 @@ export function idsToIgnoredRules(ruleIds = []) {
     return ids
   }, {})
 }
+
+export function validatePoint(textEditor, line, col) {
+  const buffer = textEditor.getBuffer()
+  // Clip the given point to a valid one, and check if it equals the original
+  if (!buffer.clipPosition([line, col]).isEqual([line, col])) {
+    throw new Error(`${line}:${col} isn't a valid point!`)
```
IanVS: I'm not convinced that throwing an error is the right thing to do here.  We could be more graceful and just use the clipped position.  I suppose the positive here is that we might be able to catch ESLint errors, whereas without throwing an error people will just live with the possibly slightly-inaccurate location marking.  Pros and cons either way, but personally I think I lean towards not throwing an error.  Curious to hear your thoughts.

src/main.js
Arcanemagus: Ooo, I like that idea.

src/main.js
Arcanemagus: That's what I get for falling in with the lazy crowd that abuses that hidden check, thanks!

src/main.js
Arcanemagus: I assumed `endColumn` is 1-based like all the other points we get from ESLint... did they really make 1/4 of the points 0-based?!?

src/helpers.js
Arcanemagus: This should be validating the point in a similar manner to how `rangeFromLineNumber()` does. Where this is being called [over here](https://github.com/AtomLinter/linter-eslint/pull/709/files/86381852cd4f81dc140230e4270a7347e9974ba1#diff-1c90ff38a08209f9ebd4d05d1e43358eR173) is wrapped in the same `try`/`catch` that the call to `rangFromLineNumber()`, so this `throw` will be caught and handled in the exact same manner as any other invalid point we run into.

src/main.js
IanVS: column numbers are 0-based, line numbers are 1-based.

Arcanemagus: Looks like _rules_ have [mixed](https://github.com/eslint/eslint/blob/0a44361d8ff319c07b92bf6c6a35890337deac0e/docs/developer-guide/working-with-rules.md#contextreport) 1-based and 0-based coords, but the results are all 1-based.

PR #716, opened 2016-09-25
Fix Failing and Flakey "Fix" Specs

IanVS:
While looking into the spec failures currently being experienced in https://github.com/AtomLinter/linter-eslint/pull/666, I noticed that some specs were failing which weren't directly related to the changes in that PR.  I found the specs to be failing for me locally as well, so this is an attempt to clean them up and get master to a good, passing state.

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -1,12 +1,15 @@
 'use babel'
 
 import * as path from 'path'
+import * as fs from 'fs'
+import { tmpdir } from 'os'
 import linter from '../lib/main'
 
 const goodPath = path.join(__dirname, 'fixtures', 'files', 'good.js')
 const badPath = path.join(__dirname, 'fixtures', 'files', 'bad.js')
 const emptyPath = path.join(__dirname, 'fixtures', 'files', 'empty.js')
 const fixPath = path.join(__dirname, 'fixtures', 'files', 'fix.js')
+const configPath = path.join(__dirname, 'fixtures', 'configs', '.eslintrc.yml')
 const importingpath = path.join(__dirname, 'fixtures',
   'import-resolution', 'nested', 'importing.js')
 const badImportPath = path.join(__dirname, 'fixtures',
@@ -19,6 +22,13 @@ describe('The eslint provider for Linter', () => {
 
   const worker = spawnWorker()
   const lint = linter.provideLinter.call(worker).lint
+  const fix = textEditor =>
+    worker.worker.request('job', {
+      contents: textEditor.getText(),
+      type: 'fix',
+      config: atom.config.get('linter-eslint'),
+      filePath: textEditor.getPath()
+    })
 
   beforeEach(() => {
     atom.config.set('linter-eslint.disableFSCache', false)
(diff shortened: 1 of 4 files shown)


IanVS:
Note, you will see in the latest [CircleCI tests on master](https://circleci.com/gh/AtomLinter/linter-eslint/621), this console log:

>Uncaught (in promise) Error: Cannot mark location in editor for (semi) - (Extra semicolon.) at line (1) column (12)", source: /home/ubuntu/linter-eslint/lib/main.js (158)

This was occurring during the spec for auto-fixing, which caused the test to not work properly but somehow also not fail.

IanVS:
Looks like Appveyor doesn't like the way I'm removing the temporary directory.  Any suggestions there?

spec/linter-eslint-spec.js
```diff
@@ -1,12 +1,17 @@
 'use babel'
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
```
Arcanemagus: That's the default value, does this line change anything?

spec/linter-eslint-spec.js
```diff
@@ -19,6 +24,13 @@ describe('The eslint provider for Linter', () => {
 
   const worker = spawnWorker()
   const lint = linter.provideLinter.call(worker).lint
+  const fix = textEditor =>
+    worker.worker.request('job', {
+      contents: textEditor.getText(),
```
Arcanemagus: `fix` jobs don't use the contents, no need to pass them in here.

spec/linter-eslint-spec.js
```diff
@@ -84,8 +96,8 @@ describe('The eslint provider for Linter', () => {
         expect(messages[0].fix.range).toEqual([[0, 11], [0, 12]])
         expect(messages[0].fix.newText).toBe('')
 
-        expect(messages[1].fix.range).toEqual([[2, 1], [2, 1]])
-        expect(messages[1].fix.newText).toBe(' ')
+        expect(messages[1].fix.range).toEqual([[2, 0], [2, 1]])
+        expect(messages[1].fix.newText).toEqual('  ')
```
Arcanemagus: `toEqual` is `==`, `toBe` is `===`, this should have stayed `toBe`, right?

spec/linter-eslint-spec.js
```diff
@@ -128,21 +140,70 @@ describe('The eslint provider for Linter', () => {
     })
   })
 
-  describe('Fix errors when saved', () => {
+  describe('Fix errors', () => {
```
Arcanemagus: "Fix" -> "Fixes"

spec/linter-eslint-spec.js
```diff
@@ -128,21 +140,70 @@ describe('The eslint provider for Linter', () => {
...
+    let doneCheckingFixes
+    let tempFixtureDir
+    let tempFixturePath
+    let tempConfigPath
+
     beforeEach(() => {
-      atom.config.set('linter-eslint.fixOnSave', true)
+      waitsForPromise(() => {
+        tempFixtureDir = fs.mkdtempSync(tmpdir() + path.sep)
+        tempFixturePath = `${tempFixtureDir}${path.sep}fixed.js`
```
Arcanemagus: Any particular reason not to use `path.join`? This works just fine as is, just the rest of the file uses `path.join`.

spec/linter-eslint-spec.js
```diff
@@ -128,21 +140,70 @@ describe('The eslint provider for Linter', () => {
...
-          })
+
+    afterEach(() => {
+      rimraf.sync(tempFixtureDir)
+    })
+
+    it('should fix linting errors', () => {
+      function firstLint(textEditor) {
+        return lint(textEditor)
+        .then((messages) => {
```
Arcanemagus: This should be indented (or just moved to the line above).

spec/linter-eslint-spec.js
```diff
@@ -128,21 +140,70 @@ describe('The eslint provider for Linter', () => {
...
+          expect(messages.length).toBe(2)
+          return textEditor
         })
+      }
+      function makeFixes(textEditor) {
+        return fix(textEditor)
+          .then((messagesAfterSave) => {
+            // Linter reports a successful fix
+            expect(messagesAfterSave).toBe('Linter-ESLint: Fix Complete')
+            return textEditor
```
Arcanemagus: Nothing is being chained after this, so this isn't necessary. It's not harming anything though so it's fine to leave in.

spec/linter-eslint-spec.js
IanVS: The default [from airbnb](https://github.com/airbnb/javascript/blob/0157e69a0df8ee8fa568d2c0ed53c087e4910c80/packages/eslint-config-airbnb-base/rules/imports.js#L69) is `"devDependencies: false"` , so it was complaining about `rimraf` not being in `dependencies` in the `package.json`.

spec/linter-eslint-spec.js
IanVS: Hah, yeah power of suggestion [(from the node docs)](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback).  `path.join` would indeed be better, I'll switch it out.

spec/linter-eslint-spec.js
Arcanemagus: Ah ha! Good to know.

IanVS:
Thanks for the thorough review, @Arcanemagus.  All excellent points, and I've pushed up a commit to address them.

spec/linter-eslint-spec.js
IanVS: To be honest, it would probably be better to make this change in a config file for the whole specs directory.  I need to run out now, but I can get to it later if you want to block the merge.  Otherwise, I'll make the change as part of a subsequent effort to get #666 passing.

spec/linter-eslint-spec.js
```diff
@@ -149,9 +148,9 @@ describe('The eslint provider for Linter', () => {
 
     beforeEach(() => {
       waitsForPromise(() => {
-        tempFixtureDir = fs.mkdtempSync(tmpdir() + path.sep)
```
Arcanemagus: Looks like the removal of this trailing separator broke the specs.

IanVS: Sigh, I can't win.

IanVS:
OK, finally passing again.  :)

Issue #680, opened 2016-08-25
Error message is totally misleading - Error: ESLint not found, Please install or make sure Atom is getting $PATH correctly


sjmcdowall:
I spent 4 hours tracking this down -- only to find out the error message is totally wrong and mis-leading.

linter-eslint doesn't even't TRY to use $PATH to find eslint -- instead it does a lot of dancing around nodePrefixPaths() etc. which can be totally wrong depending on the situation. 

For example I use "nvm" but also load common -g type modules in the "standard" /usr/local area.  From my shell I can "which eslint" and run it just fine.  I then launch atom .. but get this error message if it's set to Use Global because my nvm get prefix isn't point to the global prefix area of /usr/local but rather some nvm/v6.3.1 path thing ..

It would be awesome of complex environments to ALSO check to see if eslint DOES exist in the path and to use that if it does.  Certainly though the error message needs to change because it's just wrong and frustrating going down a blind alley. 
## 

``` js
function getESLintFromDirectory(modulesDir, config) {
  var ESLintDirectory = null;

  if (config.useGlobalEslint) {
    var prefixPath = config.globalNodePath || getNodePrefixPath();
    if (process.platform === 'win32') {
      ESLintDirectory = _path2.default.join(prefixPath, 'node_modules', 'eslint');
    } else {
      ESLintDirectory = _path2.default.join(prefixPath, 'lib', 'node_modules', 'eslint');
    }
  } else {
    ESLintDirectory = _path2.default.join(modulesDir || '', 'eslint');
  }
  try {
    return require(_path2.default.join(ESLintDirectory, 'lib', 'cli.js'));
  } catch (e) {
    if (config.useGlobalEslint && e.code === 'MODULE_NOT_FOUND') {
      throw new Error('ESLint not found, Please install or make sure Atom is getting $PATH correctly');
    }
    return require(_path2.default.join(Cache.ESLINT_LOCAL_PATH, 'lib', 'cli.js'));
  }
}
```

Arcanemagus:
Sorry, but this is a wontfix for what you _want_. We need the actual eslint JS files, not what you get from `which eslint`.

Leaving this open though as the error message could be improved.

sjmcdowall:
You then missed what _wanted_ .  What I want is an appropriate and not misleading error message, nothing else.

soredake:
Any progress on this?

IanVS:
[...]

PR #725, opened 2016-10-06
Add debug command

IanVS:
This PR adds a `Linter Eslint: Debug` command which will pop open a notification with some useful debugging information including various package versions, the type of ESLint is being used for the current file, the linter-eslint configuration, time since the last atom restart, and more.  

My intent is to add an issue template which asks users to provide the output of this command, to assist package maintainers in troubleshooting and triaging incoming issue reports.

Example output:

image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,5 +1,6 @@
 'use babel'
 
+import Path from 'path'
 import escapeHTML from 'escape-html'
 import ruleURI from 'eslint-rule-documentation'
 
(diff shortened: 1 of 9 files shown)


src/main.js
```diff
@@ -57,6 +58,36 @@ module.exports = {
     }))
 
     this.subscriptions.add(atom.commands.add('atom-text-editor', {
+      'linter-eslint:debug': () => {
+        const textEditor = atom.workspace.getActiveTextEditor()
+        const filePath = textEditor.getPath()
+        const linterEslintMeta = require(Path.join(atom.packages.resolvePackagePath('linter-eslint'), 'package.json'))
+        const config = atom.config.get('linter-eslint')
+        const configString = JSON.stringify(config, null, 2)
+        const hoursSinceRestart = process.uptime() / 3600
```
Arcanemagus: Do we want hours or minutes?

src/worker-helpers.js
```diff
@@ -29,19 +30,43 @@ export function getNodePrefixPath() {
...
+    if (fs.statSync(eslintDir).isDirectory()) {
+      return {
+        path: eslintDir,
+        type: locationType,
+      }
+    }
+  } catch (e) {
+    if (config.useGlobalEslint && e.code === 'ENOENT') {
+      throw new Error(
+          'ESLint not found, Please install or make sure Atom is getting $PATH correctly'
```
Arcanemagus: Might as well fix #680 here, maybe something along the lines of ``"Global ESLint not found, please ensure `npm get prefix` returns correctly or set the path in settings"``?

src/main.js
IanVS: I think tenths of hours (6 min resolution, which is what it is showing now) is enough to tell us whether it's been restarted recently, without being ridiculous when it hasn't been restarted in days.  But, happy to change it if you think minutes would be better.

src/worker-helpers.js
IanVS: Yeah happy to change if you like, or I can submit another PR right after this one, since it's kind of a different issue.

Arcanemagus: That's probably a better idea, just saw that what was reporting the initial message is further down in the code here anyway.

src/main.js
Arcanemagus: Wow, completely forgot to see how that looked in the output. This is just fine for me.

Arcanemagus:
Reverting this so we can get https://github.com/AtomLinter/linter-eslint/pull/609 merged.

PR #730, opened 2016-10-08
Add debug command

IanVS:
This is a second try at https://github.com/AtomLinter/linter-eslint/pull/725, which was reverted in order to merge a PR by a community member.  I rebased the code from that other PR, and am now pushing it up with a new commit.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -49,7 +49,7 @@ function spawnWorker() {
 }
 
 function showError(givenMessage) {
-  var givenDetail = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : null;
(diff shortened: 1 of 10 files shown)


lib/helpers.js
```diff
@@ -49,7 +49,7 @@ function spawnWorker() {
 }
 
 function showError(givenMessage) {
-  var givenDetail = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : null;
+  var givenDetail = arguments.length <= 1 || arguments[1] === undefined ? null : arguments[1];
```
Arcanemagus: I wish `babel` would make up its mind here... it keeps swapping the logic around.

PR #734, opened 2016-10-11
Add ESLint to the devDependencies

Arcanemagus:
As ESLint is not only a dependency of this project, but also used for developing it, it should have been listed in here from the start. Brought to the forefront by https://github.com/atom/ci/pull/60 as we discovered that linting wasn't actually running on the CI builds.


Arcanemagus:
Travis-CI seems to have forgotten about this, tried restarting the build....

PR #756, opened 2016-11-18
Improve Atom startup time

walles:
Before this change, activation was done on Atom startup, whether or not
you actually had any Javascript files open.

With this change in place, we postpone activation until the Atom
Javascript grammar's first use.

This improves startup time of my Atom by about 180ms, thus fixing one of
the top startup time offenders according to TimeCop.

For some frame of reference, I have 87 packages installed, and Timecop
lists all packages with startup times above 5ms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "engines": {
     "atom": ">=1.10.0 <2.0.0"
   },
+  "activationHooks": ["language-javascript:grammar-used"],
   "configSchema": {
     "lintHtmlFiles": {
       "title": "Lint HTML Files",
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Unfortunately I'm going to have to close this on this package, as ESLint is flexible enough that it gets used on _many_ different odd languages, so the allowed scopes (languages) are actually [configurable](https://github.com/AtomLinter/linter-eslint/blob/3b9e2dbc34f8b30fbafdcbca98382dd10c27530e/package.json#L75). This means that the `activationHook` would potentially block the package from running if somebody has configured it to work with a more esoteric language.

If you can think of a way around that though I'd love to do so!

walles:
This is more of an architectural suggestion, and I have no idea how realistic it is, but from a usability perspective all linters and probably lots of other plugins could be delayed until after the user has actually started editing.

Would it be possible to have an activation-hook like functionality that just delays activation of a package until after the user has gotten a UI where they can start editing?

Arcanemagus:
Possibly? As far as I know the only `activationHook`'s that are currently available are the `language-package-name:grammar-used` hooks.

Once https://github.com/atom/atom/pull/13200 is released to stable, that would be an interesting candidate for packages like this.

PR #761, opened 2016-12-01
Gracefully handle invalid points

### LINKED ISSUE #735: Configuration option to disable notifications
Over the last few weeks, I have been getting a lot of error notifications from stuff like https://github.com/AtomLinter/linter-eslint/issues/700. The solution in #700 (to upgrade to eslint@3.5.0) is not an option for me; I am using a legacy eslint@2.x config that would take a lot of time to upgrade to 3.5.0 because it uses many plugins. Running eslint on its own does not cause issues, and I don't want to spend a ton of time and change many lines of code due to rule incompatibilities and changes just so my editor stops annoying me.

99.9% of the time that an error has popped up from linter-eslint while I was typing, I didn't care and could safely dismiss it. It is very irritating that these ignorable errors are popping up while I type.

Since I am aware that certain rule checks fail due to race conditions on save, and I want to ignore them, it'd be nice to have an option in linter-eslint to disable all notifications from that package, or at least all error notifications.

Arcanemagus:
Gracefully handle invalid points reported by ESLint, transforming them into regular Linter messages. These messages include a link to allow easy reporting of an issue to this repository.

Example of messages in the in-development Linter UI:
image

And the generated issue:
image

Fixes #735.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -163,6 +163,15 @@ const generateInvalidTrace = async (
   }
 }
 
+/**
+ * Given a raw response from ESLint, this processes the messages into a format
+ * compatible with the Linter API.
+ * @param  {Object}     response   The raw response from ESLint
+ * @param  {TextEditor} textEditor The Atom::TextEditor of the file the messages belong to
+ * @param  {bool}       showRule   Whether to show the rule in the messages
+ * @param  {Object}     worker     The current Worker process to send Debug jobs to
+ * @return {Promise}               The messages transformed into Linter messages
+ */
 export async function processESLintMessages(response, textEditor, showRule, worker) {
   return Promise.all(response.map(async ({
     message, line, severity, ruleId, column, fix, endLine, endColumn
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -64,3 +64,49 @@ export function validatePoint(textEditor, line, col) {
     throw new Error(`${line}:${col} isn't a valid point!`)
   }
 }
+
+export async function getDebugInfo(worker) {
+  const textEditor = atom.workspace.getActiveTextEditor()
+  const filePath = textEditor.getPath()
+  const packagePath = atom.packages.resolvePackagePath('linter-eslint')
+  // eslint-disable-next-line import/no-dynamic-require
+  const linterEslintMeta = require(join(packagePath, 'package.json'))
+  const config = atom.config.get('linter-eslint')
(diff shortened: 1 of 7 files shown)


Arcanemagus:
Looks like the changes to the debugging information command broke some of the specs, investigating.

src/helpers.js
```diff
@@ -57,10 +60,177 @@ export function idsToIgnoredRules(ruleIds = []) {
...
+    let msgCol
+    let msgEndLine
+    let msgEndCol
+    let eslintFullRange = false
+    if (typeof endColumn !== 'undefined' && typeof endLine !== 'undefined') {
+      eslintFullRange = true
+      // Here we always want the column to be a number
+      msgCol = Math.max(0, column - 1)
+      msgEndLine = endLine - 1
+      msgEndCol = endColumn - 1
```
IanVS: It might be good to include a comment about why we are subtracting 1 from a lot of these values, as it isn't immediately obvious.

src/helpers.js
```diff
@@ -57,10 +60,177 @@ export function idsToIgnoredRules(ruleIds = []) {
...
+        type: 'Trace',
+        text: rangeText,
+        filePath,
+        severity: 'info',
+      },
+    ]
+  }
+}
+
+export async function processMessages(response, textEditor, showRule, worker) {
```
IanVS: Since this is a pretty generic function name, it might be worthwhile to add a short description of what it does in a comment.  Or better yet in a doc block, although we haven't been good about adding them up until now.

src/helpers.js
```diff
@@ -57,10 +60,177 @@ export function idsToIgnoredRules(ruleIds = []) {
...
+        ret.html = `<${elName}${href} class="badge badge-flexible eslint">` +
+          `${ruleId || 'Fatal'}</${elName}> ${escapeHTML(message)}`
+      } else {
+        ret.text = message
+      }
+      if (linterFix) {
+        ret.fix = linterFix
+      }
+    } catch (err) {
+      ret = await generateInvalidTrace(
```
IanVS: Are you pretty sure that the only kinds of errors which will be caught will be from invalid ranges?  If not, it would be good to check the `err` and re-throw if it is not due to an invalid point.

Arcanemagus: Thanks for reminding me, I'd thought about that during development but forgot to actually implement it in the middle of getting this all working 😛.

Arcanemagus:
@IanVS updated 😉.

PR #777, opened 2017-01-05
Determine fix results based on exit code

### LINKED ISSUE #511: Notification expecting string but object is passed
While fixing errors with `Fix error` option , for some reason the fix failed . But notification received is an object which we are passing as is to atom notification . Which is expecting a string , so it does not show the error message in notification. 
I get the following error in console 

```
Uncaught (in promise) Error: Notification must be created with string message: Error: Linter-ESLint: Fix Attempt Completed, Linting Errors Remain(…)
```

After debugging I realized the `response` object is of the form 

``` js
{
   "message" : "Linter-ESLint: Fix Attempt Completed, Linting Errors Remain"
}
```

The code in `linter-eslint` is like this : 

``` js
        this.worker.request('job', {
          type: 'fix',
          config: atom.config.get('linter-eslint'),
          filePath
        }).then((response) =>
          atom.notifications.addSuccess(response)
        ).catch((response) =>
          atom.notifications.addWarning(response)
        )

```

I am not sure if this something that should be fixed here or else where .

IanVS:
Fixes #511 

## Description

This adjusts the response of running `Linter eslint: fix` or fixing automatically on save, so that unexpected runtime errors are correctly printed in a notification.

This PR also no longer treats an incomplete `fix` as requiring a warning.  A success message is given whether the fix is total or partial, with only the wording being altered to reflect the result.  As I mention in the commit message details, I believe `eslint.execute` used to throw when autofix did not fix all of the errors in a file.  It no longer does, so we need to rely on the exit code instead.

---
This change is [Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/777)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -14,6 +14,7 @@ import { getConfigPath } from './worker-helpers'
 const scopes = []
 let showRule
 let ignoredRulesWhenModified
+let disableWhenNoEslintConfig
 
 module.exports = {
   activate() {
@@ -53,7 +54,7 @@ module.exports = {
           // Do not try to fix if linting should be disabled
           const fileDir = Path.dirname(filePath)
           const configPath = getConfigPath(fileDir)
-          if (configPath === null && atom.config.get('linter-eslint.disableWhenNoEslintConfig')) return
+          if (configPath === null && disableWhenNoEslintConfig) return
 
           this.worker.request('job', {
             type: 'fix',
@@ -103,6 +104,12 @@ module.exports = {
     this.subscriptions.add(
       atom.config.observe('linter-eslint.showRuleIdInMessage', (value) => {
         showRule = value
+      })
+    )
+
+    this.subscriptions.add(
+      atom.config.observe('linter-eslint.disableWhenNoEslintConfig', (value) => {
+        disableWhenNoEslintConfig = value
       })
     )
 
--- a/src/main.js
+++ b/src/main.js
@@ -53,9 +53,9 @@ module.exports = {
             config: atom.config.get('linter-eslint'),
             filePath,
             projectPath
-          }).catch(response =>
-            atom.notifications.addWarning(response)
-          )
+          }).catch((err) => {
+            atom.notifications.addWarning(err.message)
+          })
         }
       })
     }))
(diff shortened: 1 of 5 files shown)


IanVS:
Hm, I just noticed that this change will cause `linter-eslint` to display a warning of: `No ESLint configuration found.` if `Fix errors on save` is checked or a manual `fix` is run, even if `Disable when no ESLint config is found` is checked.

It's nice to get the warning in the manual case, but I think saving a file should not show that message if the `Disable` option is checked.  I'll need to work on that.

IanVS:
OK, the latest commit should help.  We will no longer attempt to perform a `fix` job when `disableWhenNoEslintConfig` is true and there's no config file found.  But, we will still attempt it if a manual `Linter Eslint: fix` command is run, and a helpful message of `No ESLint configuration found.` is shown.

src/main.js
```diff
@@ -48,14 +50,19 @@ module.exports = {
           const filePath = editor.getPath()
           const projectPath = atom.project.relativizePath(filePath)[0]
 
+          // Do not try to fix if linting should be disabled
+          const fileDir = Path.dirname(filePath)
+          const configPath = getConfigPath(fileDir)
+          if (configPath === null && atom.config.get('linter-eslint.disableWhenNoEslintConfig')) return
```
Arcanemagus: This setting should be observed instead of retrieved every time.

src/worker.js
```diff
@@ -33,12 +33,11 @@ function lintJob(argv, contents, eslint, configPath, config) {
...
-  try {
-    eslint.execute(argv)
-    return 'Linter-ESLint: Fix Complete'
-  } catch (err) {
-    throw new Error('Linter-ESLint: Fix Attempt Completed, Linting Errors Remain')
+  const exit = eslint.execute(argv)
+  if (exit === 0) {
+    return 'Linter-ESLint: Fix complete.'
   }
+  return 'Linter-ESLint: Fix attempt complete, but linting errors remain.'
```
IanVS: Actually, to get an error code other than `0`, means there are errors.  If there are only warnings, eslint will exit with a `0`.

Arcanemagus: Ah, didn't know that. `errors` works just fine then!

IanVS:
OK, @Arcanemagus, I made the fix you asked for, and rebased on master.  I think this is ready-to-go.

PR #778, opened 2017-01-09
Ignore home config from disableWhenNoEslintConfig

### LINKED ISSUE #773: Do not consider configs in `~/` for `Disable when no Eslint Config is found` setting
## Issue Type
Bug on version 8.0.0

## Issue Description
In my project I have no package.json nor a .eslintrc and when saving an .js file an error prompt about `Error: Failed to load plugin html: Cannot find module 'eslint-plugin-html` rise.

I just want to use nasty js code for this project, why this plugin cares so much!? :)

Let me know if you need any additional information.

IanVS:
Fixes #773

## Description

This PR prevents an eslint config file located in a user's home directory (e.g. `/users/person/.eslintrc`) from enabling linting of projects within the home directory when the setting `Disable when no ESLint config is found` is enabled.  

Note, because of the way we traverse up the directory structure looking for config files, projects outside of a user's home do not currently find a `fallback` config in a user's home, and thus those projects are already being disabled correctly.

## Notes

This adds a dependency on `user-home`, which is used to detect the user's home directory and return it as a string.  This result is compared to the directory containing the config file returned from the config file search, to determine if the file should be treated as a valid project config file or a fallback.

Also, I've added a few specs here around the `disableWhenNoEslintConfig` setting, but I wasn't able to create a good red-green test for the particular fix in this PR.  That's because I could not find a way to mock the functionality of `findCached` in such a way to fake the finding of a config in the home directory.  My plan was to put a file with linting errors in the OS temp dir, (not under the home dir), mock out `findCached` to make it seem there was a config file in the user home, and then show that with this PR, linting is still disabled.  However, my go-to tool for such purposes, `proxyquire` does not seem to work correctly with electron's `require` implementation.  If you know if any other way to mock dependencies in the way I'm attempting, I'd be keen to hear it.

## How to test
Despite not having a good automated test for this change, it's fairly easy to check manually.

- Put a `.js` file similar to the `badInline.js` file added in this pr, somewhere in your home directory, like `~/experiment/foo.js`.
- Assuming you don't already have an eslint config in your home, `touch ~/.eslintrc`
- Verify that without the changes in this PR, the setting for `disableWhenNoEslintConfig` has no effect, and linting is performed on the file no matter the setting.
- Then, with the changes in this PR, verify that linting is indeed disabled when (and only when) `disableWhenNoEslintConfig` is turned on.

## Warning

The base branch of this PR is not master, but rather another branch with an open PR.  I have a good feeling that other PR (#777) will be merged, and rather than rebase this branch afterwards, I've just based on that, and will move the base to master after #777 is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/atomlinter/linter-eslint/778)

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -10,6 +10,7 @@ const fixturesDir = path.join(__dirname, 'fixtures')
 
 const goodPath = path.join(fixturesDir, 'files', 'good.js')
 const badPath = path.join(fixturesDir, 'files', 'bad.js')
+const badInlinePath = path.join(fixturesDir, 'files', 'badInline.js')
 const emptyPath = path.join(fixturesDir, 'files', 'empty.js')
 const fixPath = path.join(fixturesDir, 'files', 'fix.js')
 const configPath = path.join(fixturesDir, 'configs', '.eslintrc.yml')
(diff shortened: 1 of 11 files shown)


spec/linter-eslint-spec.js
```diff
@@ -361,4 +362,94 @@ describe('The eslint provider for Linter', () => {
...
+
+  describe('when `disableWhenNoEslintConfig` is true', () => {
+    let editor
+    let tempFixtureDir
+    let tempFixturePath
+    beforeEach(() => {
+      atom.config.set('linter-eslint.disableWhenNoEslintConfig', true)
+
+      waitsForPromise(() =>
+        new Promise((resolve) => {
```
Arcanemagus: Since this functionality is being used twice maybe it should be split into a function?

src/helpers.js
```diff
@@ -255,3 +256,16 @@ export async function processESLintMessages(response, textEditor, showRule, work
...
+/**
+ * Check if a config is directly inside a user's home directory.
+ * Such config files are used by ESLint as a fallback, only for situations
+ * when there is no other config file between a file being linted and root.
+ *
+ * @param  {string}  configPath - The path of the config file being checked
+ * @return {Boolean}            [description]
+ */
+export function isConfigAtHomeRoot(configPath) {
+  console.log(userHome, dirname(configPath);
```
Arcanemagus: Looks like this debug statement got left in 😛.

src/main.js
```diff
@@ -54,7 +54,8 @@ module.exports = {
           // Do not try to fix if linting should be disabled
           const fileDir = Path.dirname(filePath)
           const configPath = getConfigPath(fileDir)
-          if (configPath === null && disableWhenNoEslintConfig) return
+          const noProjectConfig = (configPath === null || isConfigAtHomeRoot(configPath))
+          if (noProjectConfig && disableWhenNoEslintConfig) return
```
Arcanemagus: I'm personally not a fan of one line `if` statements, but I'll leave that up to you.

src/worker.js
```diff
@@ -6,6 +6,7 @@ import Path from 'path'
 import { create } from 'process-communication'
 import { FindCache, findCached } from 'atom-linter'
 import * as Helpers from './worker-helpers'
+import { isConfigAtHomeRoot } from './helpers'
```
Arcanemagus: `helpers` brings in a _lot_ of things, and isn't intended to be used from the worker process (gotta love our convoluted design here...). Maybe this function should live in `worker-helpers` and helpers imports it from there? Either way is messy 😕.

src/helpers.js
IanVS: whoops.

src/main.js
IanVS: That's how it was before, so I'm just going to leave it for now.  :)

IanVS:
@Arcanemagus I believe I've addressed your feedback.

spec/linter-eslint-spec.js
```diff
@@ -367,24 +380,14 @@ describe('The eslint provider for Linter', () => {
     let editor
     let didError
     let gotLintingErrors
-    let tempFixtureDir
-    let tempFixturePath
+    const tempFixtureDir = fs.mkdtempSync(tmpdir() + path.sep)
```
Arcanemagus: Since this is the exact same in both test cases, you _could_ move it into `copyFileToTempDir`, if you want to leave it as a parameter though you might as well just call the function `copyFileToDir` 😛.

IanVS: I passed it in so that I could rimraf it after the test. Guess I could indeed rename it.

IanVS: @Arcanemagus I ended up moving it to the function after all.  Seemed … cleaner, I guess.

PR #811, opened 2017-02-10
Use cursor scope when checking fixOnSave scopes

### LINKED ISSUE #794: html fix errors on save not working
## Issue Type

Bug

## Issue Description

"Fix errors on save" not working with .html in  <script></script> tags.

"Fix errors on save" works with .js files.
Linting on html works(plugin for html installed).
.eslintrc:
```
{
  "plugins": ["html"],
  "extends": "airbnb"
}
```
Original code:
```JavaScript
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <title>👀</title>
    <link rel="stylesheet" href="style.css">
  </head>
  <body>
    <script>
    function someFunc() {
      return arr.filter(place => {
        //some code
      });
    }
    </script>
  </body>
</html>
```

Expected result in <script></script>:
```JavaScript
function someFunc() {
  return arr.filter((place) => {  // added () around "place"
    // some code    <= // added space
  });
}
```

## Bug Checklist
- [x] Restart Atom
- [x] Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not

```
Linter Eslint: 
atom version: 1.13.0
linter-eslint version: 8.0.0
eslint version: 3.14.1
hours since last atom restart: 0.3
platform: linux
Using global eslint from /home/user/.node/lib/node_modules/eslint
  "fixOnSave": true,
  "lintHtmlFiles": true,
  "useGlobalEslint": true,
  "showRuleIdInMessage": true,
  "eslintrcPath": "",
  "globalNodePath": "",
  "eslintRulesDir": "",
  "disableEslintIgnore": false,
  "disableFSCache": false,
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic"
  ],
  "rulesToSilenceWhileTyping": []
}
```

Arcanemagus:
When the fixOnSave callback is ran previously we were using the file level "master scope", this fails though when the user has enabled HTML embedded JS linting. Use the cursors of the provided editor to check the valid scopes.

Fixes #794.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -54,8 +54,7 @@ module.exports = {
 
     this.subscriptions.add(atom.workspace.observeTextEditors(editor => {
       editor.onDidSave(() => {
-        const editorScopes = editor.getLastCursor().getScopeDescriptor().scopes;
-        const validScope = editorScopes.some(scope => scopes.includes(scope));
+        const validScope = editor.getCursors().some(cursor => cursor.getScopeDescriptor().getScopesArray().some(scope => scopes.includes(scope)));
         if (validScope && atom.config.get('linter-eslint.fixOnSave')) {
           const filePath = editor.getPath();
           const projectPath = atom.project.relativizePath(filePath)[0];
--- a/lib/main.js
+++ b/lib/main.js
@@ -54,7 +54,9 @@ module.exports = {
 
     this.subscriptions.add(atom.workspace.observeTextEditors(editor => {
       editor.onDidSave(() => {
-        if (scopes.indexOf(editor.getGrammar().scopeName) !== -1 && atom.config.get('linter-eslint.fixOnSave')) {
+        const editorScopes = editor.getLastCursor().getScopeDescriptor().scopes;
+        const validScope = editorScopes.some(scope => scopes.includes(scope));
+        if (validScope && atom.config.get('linter-eslint.fixOnSave')) {
           const filePath = editor.getPath();
           const projectPath = atom.project.relativizePath(filePath)[0];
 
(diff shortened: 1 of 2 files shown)


lib/main.js
```diff
@@ -54,7 +54,9 @@ module.exports = {
 
     this.subscriptions.add(atom.workspace.observeTextEditors(editor => {
       editor.onDidSave(() => {
-        if (scopes.indexOf(editor.getGrammar().scopeName) !== -1 && atom.config.get('linter-eslint.fixOnSave')) {
+        const editorScopes = editor.getLastCursor().getScopeDescriptor().scopes;
+        const validScope = editorScopes.some(scope => scopes.includes(scope));
```
IanVS: I don't know the performance impact of checking the scope, but we could probably first check that `fixOnSave` is true before calculating that the scope is valid, right?

Arcanemagus: Actually since currently this code is using an `atom.config.get()` call this is _much_ cheaper than checking that setting value since that has to do a ton of resolving and checking of things before it returns a value. When that gets moved to an `observe` in a future code overhaul then it will be better to check it first.

src/main.js
```diff
@@ -48,8 +48,10 @@ module.exports = {
 
     this.subscriptions.add(atom.workspace.observeTextEditors((editor) => {
       editor.onDidSave(() => {
-        if (scopes.indexOf(editor.getGrammar().scopeName) !== -1 &&
-            atom.config.get('linter-eslint.fixOnSave')) {
+        const validScope = editor.getCursors().some(cursor =>
+          cursor.getScopeDescriptor().some(scope =>
```
Arcanemagus: Whoops, this returns a `ScopeDescriptor`, not the Array of scopes.

PR #843, opened 2017-03-08
Don't specify rulesDir if not found

### LINKED ISSUE #759: Rules dir should be optional
## Issue Type
Feature Request

## Issue Description
I've been working on a project which uses some extra rules defined in a directory as part of the code. If I want Atom to pick them up, I need to set the path to that directory in the plugin.

However, when I work in other projects the plugin does nothing because it doesn't find that directory.

It would be nice to have that being optional, so if it's not detected it just ignores it.

fcFn:
Fixes #759.


Arcanemagus:
Released in v8.1.3. 🎉

PR #853, opened 2017-03-17
restore cursor position after fix command

yacut:
The cursor at the moment jumps to the beginning of the file after the fix command. It infuriates me :)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -70,12 +70,15 @@ module.exports = {
             rules = ignoredRulesWhenFixing;
           }
 
+          const cursorPosition = editor.getCursorBufferPosition();
           this.worker.request('job', {
             type: 'fix',
             config: atom.config.get('linter-eslint'),
             rules,
             filePath,
             projectPath
+          }).then(() => {
+            editor.setCursorBufferPosition(cursorPosition);
           }).catch(err => {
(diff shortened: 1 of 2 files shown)


IanVS:
Thanks for the contribution, @yacut!

yacut:
@IanVS, could you please publish the changes?

IanVS:
Actually, I've been using `master` lately which includes this change, and it does not seem to have fixed the underlying problem, unfortunately.  We may need to revert it back out, since it does not have the intended effect.  Before we do that, it would be great if I could get confirmation from someone else that this does not solve the cursor problem.  

To do that, uninstall `linter-eslint` from atom, clone this repo, then `apm link` from inside the repo.  Then use it for a while and see if the cursor still jumps to the start of the file occationally when fixing on save.

yacut:
I already installed it from the master and it's works works for me: https://giphy.com/gifs/linter-eslint-3o7bubhh3ZUgFe41qg/fullscreen
The cursor stays in the current position after the autofix.

IanVS:
It usually works for me too, the cursor resetting is intermittent.  Can you leave it on for several days and report back whether it _never_ resets.  It happens once or twice a day for me, usually.

Issue #849, opened 2017-03-15
Multiline error messages with linter 2.0 are broken


caesarsol:
(From https://github.com/steelbrain/linter/issues/1352)

## Issue Type
Bug

## Issue Description

Multiline error messages, such as Parser errors, are displayed in a single line and made unreadable.

<img width="1053" alt="screen shot 2017-03-14 at 12 48 37 pm" src="https://cloud.githubusercontent.com/assets/1799710/23899423/5ab9a45a-08b5-11e7-9c20-70c888d8712e.png">

### Manual eslint run

```bash
$ node node_modules/eslint/bin/eslint.js --cache --cache-location=/tmp/ src/

/Users/caesarsol/dev/project/file.js
  102:32  error  Parsing error: Unexpected token

  100 |     },
  101 |
> 102 |     get eventGroupsVisible() {;:;
      |                                ^
  103 |       const fromTs = this.eventsTimeExtent.max - (config.intervalToShow + config.intervalToFetch * 2)
  104 |       const toTs = this.eventsTimeExtent.max - 1
  105 |       const eventsGrouped = this.getEventsFromTo(fromTs, toTs)

✖ 1 problem (1 error, 0 warnings)
```

## Linter debug (stripped)

Platform: darwin
Atom Version: 1.15.0
Linter Version: 2.0.0
All Linter Providers: 
  - ESLint
Matching Linter Providers: 
  - ESLint
Disabled Linter Providers; 

### Other useful info

linter-eslint version: 8.1.3
Hours since last Atom restart: 0.1
Using local project ESLint from: /Users/caesarsol/dev/project/node_modules/eslint

Hope these info were useful.
Thanks for your work on this project!

## Bug Checklist
- Restart Atom
- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- Paste the output of the `Linter Eslint: Debug` command from the Command Palette below

tommai78101:
Chiming in to offer a suggestion on how to improve the design for this bug:

1. Given that when using the linter in Atom, the linter already highlights where the errors are in the codes, there is no point in using ASCII text to point out where or what line of code is causing the error.
2. It may be possible to not parse the error message after the imporant message has been stated.For example, the error message is: `Parsing error: Unexpected token, expected , 21 | 22 | var foo = { > 23 | var bar = { | ^ 24 | var baz = { 25 | 26 | }; (Fatal)`

[...]

PR #855, opened 2017-03-22
Linter Message v2

zcei:
As the `Linter Message v2` API was finalized & [released about a week ago](http://steelbrain.me/2017/03/13/linter-v2-released.html) I was very eager to try it out.

The same thing @caesarsol describes in #849 applied to me as well, and while digging into it, I found it useful to migrate `linter-eslint` to the latest API.

**This PR is not ready to be merged**, I just wanted to get input by a maintainer about how to proceed from here on.
~For example I currently didn't commit `lib`, as it somehow included comments to a file I didn't touch.~ Manually added only the compilation diff without comments. Still: *what is the intended usage?*

## (potential) TODOs

* [ ] fetch rules from ESLint page to display in new `description` field (instead of link only)
* [ ] write specs for `generateInvalidTrace` (I assume this is the render function for #849)
* [ ] think about refactoring to `solutions` (left `fix` in code, as this is the ESLint option)


Arcanemagus:
`generateInvalidTrace` is used when ESLint feeds us a point that doesn't exist in the file, so we generate a message prompting the user to report it so it can be investigated and fixed in ESLint 😉.

src/helpers.js
```diff
@@ -240,20 +234,23 @@ export async function processESLintMessages(response, textEditor, showRule, work
...
+        }
       }
 
       if (showRule) {
-        const elName = ruleId ? 'a' : 'span'
-        const href = ruleId ? ` href="${ruleURI(ruleId).url}"` : ''
-        ret.html = `${escapeHTML(message)} (<${elName}${href}>${ruleId || 'Fatal'}</${elName}>)`
+        const url = ruleId ? ruleURI(ruleId).url : ''
+        // TODO: fetch rule Markdown from eslint
+        ret.url = url
```
Arcanemagus: Since Linter now has built in handling of URL references, this should _always_ be specified if it is available. The `showRule` option should now only control whether the ` (rule-name)` text is added to the `excerpt`.

src/helpers.js
```diff
@@ -240,20 +234,23 @@ export async function processESLintMessages(response, textEditor, showRule, work
...
+        ret.description = `${escapeHTML(message)} ([${ruleId || 'Fatal'}](${url}))`
```
Arcanemagus: This should be `ret.excerpt`. `description` is only shown on expanding the message and is only intended for blocks of descriptive text, `excerpt` is the equivalent of the old `text/html` attribute.

src/helpers.js
```diff
@@ -240,20 +234,23 @@ export async function processESLintMessages(response, textEditor, showRule, work
...
       } else {
-        ret.text = message
+        ret.description = message
```
Arcanemagus: Same as above, this should be `ret.excerpt`.

src/helpers.js
Arcanemagus: Also, HTML is never rendered in Linter messages now, so `escapeHTML` can be entirely removed. (`description` allows Markdown, that's the only templating allowed.)

src/helpers.js
```diff
@@ -155,28 +155,22 @@ const generateInvalidTrace = async (
...
-        severity: 'info',
-      },
-      {
-        type: 'Trace',
-        text: rangeText,
-        filePath,
-        severity: 'info',
-      },
-    ]
+    excerpt: `${escapeHTML(titleText)}. See the trace for details. ` +
```
Arcanemagus: `escapeHTML` isn't necessary. Also there is no such thing as a trace any more 😉.

src/helpers.js
```diff
@@ -155,28 +155,22 @@ const generateInvalidTrace = async (
...
+      `[Report this!](${newIssueURL})`,
```
Arcanemagus: `newIssueURL` should be set as the `url` attribute. The message should be updated to instruct the user to click that.

src/helpers.js
```diff
@@ -155,28 +155,22 @@ const generateInvalidTrace = async (
...
+    location,
+    description: rangeText,
+    reference: {
```
Arcanemagus: This isn't really going to be useful at all any more.

src/helpers.js
```diff
@@ -155,28 +155,22 @@ const generateInvalidTrace = async (
...
-        type: 'Trace',
-        text: rangeText,
-        filePath,
-        severity: 'info',
-      },
-    ]
+    excerpt: `${escapeHTML(titleText)}. See the trace for details. ` +
+      `[Report this!](${newIssueURL})`,
+    location,
+    description: rangeText,
```
Arcanemagus: This should be a (sensible) merger of `rangeText` and the original message, allowing the user to see what rule and message ESLint _attempted_ to show, and where it requested to be shown.

Arcanemagus:
`lib` should always include the current compiled output, if that's missing a comment it means an earlier PR was merged accidentally without it 😛.

`generateInvalidTrace` would be rather annoying to write specs for as it currently isn't exported (and shouldn't be as it's an internal method in that file only used in `processESLintMessages`).

I'm not sure what you are talking about with the solutions since you already handled that?

Arcanemagus:
I suspect the issue in #849 is a parsing issue with the special format of those `parse error` messages, but haven't had time to look into it.

Arcanemagus:
Also don't forget to mark the `linter` provided service as v2 😉.

caesarsol:
@zcei thanks for tackling the bug, let me know if you need any help!

zcei:
> Also don't forget to mark the linter provided service as v2 😉.

I was pretty sure I did so 😓 

> I'm not sure what you are talking about with the solutions since you already handled that?

Question is just about naming within code - but feels like you're okay with sticking to `fix` internally, as this is the CLI arg for ESLint.

> `generateInvalidTrace` would be rather annoying to write specs for as it currently isn't exported (and shouldn't be as it's an internal method in that file only used in `processESLintMessages`).

Then I miss-understood the purpose of the fn 👍 

@Arcanemagus @caesarsol anyone of you maybe has some input in how to properly link the package into Atom? I used `apm develop` & `apm link` but neither of both seemed to work.

zcei:
@Arcanemagus the AppVeyor build is failing with some strange permission thing.

Does it really come from bumping the deps, that now something doesn't work on Windows? (POSIX user over here 😅 )

steelbrain:
Hey guys, sorry I'm a little late to the party, I've been maintaining a rather old but working ESLint fork for Linter v2 (the one I demoed in the sneak peek video), You can find it at https://github.com/steelbrain/linter-eslint/tree/steelbrain/linter-v2

spec/.eslintrc.js
```diff
@@ -2,5 +2,8 @@ module.exports = {
   env: {
     jasmine: true,
     atomtest: true
+  },
+  rules: {
```
Arcanemagus: Just remove this entire section. First of all current ESLint configurations should use the string representation of severity (`2` -> `"error"`) so it's easier to understand when reading it. Secondly the only thing this changes from the [current configuration](https://github.com/airbnb/javascript/blob/fa0a82373eb46ff8e99ccc43ba80c2c18c319e01/packages/eslint-config-airbnb-base/rules/style.js#L344) is the adding of `allowTemplateLiterals`. Currently those are already allowed _when needed_, and disallowed when they are pointless. This change here allows them to be used when they aren't needed which is worse than pointless as template strings require extra processing.

spec/linter-eslint-spec.js
```diff
@@ -141,17 +143,19 @@ describe('The eslint provider for Linter', () => {
       waitsForPromise(() =>
         atom.workspace.open(badImportPath).then(editor =>
           lint(editor).then((messages) => {
-            const expected = 'Unable to resolve path to module &#39;../nonexistent&#39;. ' +
-              '(<a href="https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md">' +
-              'import/no-unresolved</a>)'
+            const expected = `Unable to resolve path to module '../nonexistent'. ` +
+              '(import/no-unresolved)'
+
```
Arcanemagus: Remove this blank line please.

spec/linter-eslint-spec.js
```diff
@@ -141,17 +143,19 @@ describe('The eslint provider for Linter', () => {
...
+            const expectedUrl = 'https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md'
 
             expect(messages.length).toBeGreaterThan(0)
-            expect(messages[0].type).toBe('Error')
+            expect(messages[0].severity).toBe('error')
             expect(messages[0].text).not.toBeDefined()
```
Arcanemagus: Remove this, `linter` will validate this for us for v2 providers.

spec/linter-eslint-spec.js
```diff
@@ -141,17 +143,19 @@ describe('The eslint provider for Linter', () => {
...
-            expect(messages[0].html).toBe(expected)
-            expect(messages[0].filePath).toBe(badImportPath)
-            expect(messages[0].range).toEqual([[0, 24], [0, 39]])
-            expect(Object.hasOwnProperty.call(messages[0], 'fix')).toBeFalsy()
+            expect(messages[0].excerpt).toBe(expected)
+            expect(messages[0].url).toBe(expectedUrl)
+            expect(messages[0].location.file).toBe(badImportPath)
+            expect(messages[0].location.position).toEqual([[0, 24], [0, 39]])
+            expect(Object.hasOwnProperty.call(messages[0], 'solutions')).toBeFalsy()
```
Arcanemagus: Let's go with the Jasmine way:
```js
expect(messages[0].solutions).not.toBeDefined()
```

spec/linter-eslint-spec.js
```diff
@@ -269,7 +273,7 @@ describe('The eslint provider for Linter', () => {
             if (messagesAfterFixing) {
               // There is still one linting error, `semi` which was disabled during fixing
               expect(messagesAfterFixing.length).toBe(1)
-              expect(messagesAfterFixing[0].html).toBe('Extra semicolon. (<a href="http://eslint.org/docs/rules/semi">semi</a>)')
+              expect(messagesAfterFixing[0].html).toBe('Extra semicolon. ([semi](http://eslint.org/docs/rules/semi))')
```
Arcanemagus: I suspect the only reason that this didn't cause the build to fail is that this spec is rather flakey currently. This definitely should be checking against the v2 message format.

spec/linter-eslint-spec.js
```diff
@@ -287,19 +291,21 @@ describe('The eslint provider for Linter', () => {
...
+
     it('does nothing on saved files', () => {
       atom.config.set('linter-eslint.rulesToSilenceWhileTyping', ['no-trailing-spaces'])
       waitsForPromise(() =>
         atom.workspace.open(modifiedIgnoreSpacePath).then(editor =>
           lint(editor).then((messages) => {
             expect(messages.length).toBe(1)
-            expect(messages[0].type).toBe('Error')
+            expect(messages[0].severity).toBe('error')
             expect(messages[0].text).not.toBeDefined()
```
Arcanemagus: Linter will verify this for us, remove this check.

spec/linter-eslint-spec.js
```diff
@@ -316,11 +322,12 @@ describe('The eslint provider for Linter', () => {
               if (messages) {
                 // Verify the space is showing an error
                 expect(messages.length).toBe(1)
-                expect(messages[0].type).toBe('Error')
+                expect(messages[0].severity).toBe('error')
                 expect(messages[0].text).not.toBeDefined()
```
Arcanemagus: Linter will verify this for us, remove this check.

spec/linter-eslint-spec.js
```diff
@@ -411,13 +418,14 @@ describe('The eslint provider for Linter', () => {
     waitsForPromise(() =>
       atom.workspace.open(endRangePath).then(editor =>
         lint(editor).then((messages) => {
-          const expected = 'Unreachable code. ' +
-            '(<a href="http://eslint.org/docs/rules/no-unreachable">no-unreachable</a>)'
-          expect(messages[0].type).toBe('Error')
+          const expected = 'Unreachable code. (no-unreachable)'
+          const expectedUrl = 'http://eslint.org/docs/rules/no-unreachable'
+          expect(messages[0].severity).toBe('error')
```
Arcanemagus: Can you add a space between these definitions and the checks?

spec/linter-eslint-spec.js
```diff
@@ -411,13 +418,14 @@ describe('The eslint provider for Linter', () => {
...
           expect(messages[0].text).not.toBeDefined()
```
Arcanemagus: Linter will verify this for us, remove this check.

spec/linter-eslint-spec.js
```diff
@@ -450,10 +458,11 @@ describe('The eslint provider for Linter', () => {
         .then((messages) => {
           // Older versions of ESLint will report an error
           // (or if current user running tests has a config in their home directory)
-          const expectedHtml = '&#39;foo&#39; is not defined. ' +
-            '(<a href="http://eslint.org/docs/rules/no-undef">no-undef</a>)'
+          const expected = `'foo' is not defined. (no-undef)`
+          const expectedUrl = 'http://eslint.org/docs/rules/no-undef'
```
Arcanemagus: A line between this and the `expect` calls please, thanks!

spec/.eslintrc.js
Arcanemagus: Note that there are several cases below that this will affect as the strings have `'` in them, just use `"foo 'bar'"`.

src/helpers.js
```diff
@@ -240,20 +231,23 @@ export async function processESLintMessages(response, textEditor, showRule, work
...
-      if (showRule) {
-        const elName = ruleId ? 'a' : 'span'
-        const href = ruleId ? ` href="${ruleURI(ruleId).url}"` : ''
-        ret.html = `${escapeHTML(message)} (<${elName}${href}>${ruleId || 'Fatal'}</${elName}>)`
-      } else {
-        ret.text = message
-      }
+      // Fallback to ruleId for triggering a Google search
+      // Useful for custom rules, that are not matched by `ruleURI`
+      const url = ruleId ? ruleURI(ruleId).url : ruleId
```
Arcanemagus: Although this is currently the behavior of `linter`, it's not documented and is slated for removal. If providers are giving a `url` field, it _should_ be a full URL.

I would much prefer the current behavior which defaults to [this document](https://github.com/jfmengels/eslint-rule-documentation/blob/master/contributing.md), leading people on how to add the documentation. If they just want to search they can copy/paste from the panel.

If @IanVS want's though I would be okay with checking the result from `ruleURI(ruleID)`, and if `found` is `false` then we simply don't specify a `url` attribute.

steelbrain:
Another thing @Arcanemagus @zcei 

I noticed the current syntax of the excerpt is `message (rule)`. I did this a little differently in my fork mentioned above as `rule: message` which looks  a little better since you know what the problem is at the very first glance. It's my own personal taste tho, so you're free to use whichever format floats your boat.

Arcanemagus:
@steelbrain Remember that the rule is optional, I'd rather leave it on the end so if enabled it's appended on the end and if it isn't enabled then it just simply isn't there.

It's definitely personal preference though.

zcei:
Sorry was a bit busy, and at some point lost the track. Last thing I remember was adjusting with @steelbrain's branch 🤔 

Cool that you push it further, I'm happy to this this getting released! 👍 
(and thanks for the small credit 🙂 )

Arcanemagus:
Released in v8.2.0 🎉.

PR #859, opened 2017-03-23
Adding support for a comma separated list of advancedLocalNodeModules

GreenGremlin:
Different projects may have eslint installed in different locations. This change allows a user to specify a list of node_modules paths to check for eslint.

i. e. "node_modules,react-scripts/node_modules" would first look for eslint in the local node_modules, if it did not find it there it would check for react-scripts/node_modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     },
     "advancedLocalNodeModules": {
       "title": "Path to the local node_modules folder",
-      "description": "Optionally specify the path to the local node_modules folder",
+      "description": "Optionally specify the path to the local node_modules folder (separate multiple paths with a comma)",
       "type": "string",
       "default": ""
     },
--- a/lib/main.js
+++ b/lib/main.js
@@ -70,6 +70,8 @@ module.exports = {
             rules = ignoredRulesWhenFixing;
           }
 
+          // The fix replaces the file content and the cursor jumps automatically
+          // to the beginning of the file, so save current cursor position
           const cursorPosition = editor.getCursorBufferPosition();
           this.worker.request('job', {
             type: 'fix',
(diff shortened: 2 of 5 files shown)


GreenGremlin:
This helps with [create-react-app #866](https://github.com/facebookincubator/create-react-app/issues/866).

GreenGremlin:
It would be nice to see this merged, if possible.

GreenGremlin:
Could someone please let me know where this PR stands? Is there anything I can do to help get it merged?

Arcanemagus:
Looking over this now, sorry for the delay!

package.json
```diff
@@ -45,7 +45,7 @@
     },
     "advancedLocalNodeModules": {
```
Arcanemagus: This is going to be messy, but it's better for future code and better for users:

Let's remove this current config and create a new one ("overrideNodeModulesPaths"?).

This new setting will be an array type, with a schema something like this:
```json
"overrideNodeModulesPaths": {
  "title": "Override the path to the local node_modules folder",
  "description": "Multiple override paths can be separated with a comma.",
  "type": "array",
  "default": [],
  "items": {
    "type": "string"
  }
}
```

Then in the `activate()` section you'll need to migrate current users of the setting over to this new config, similar to [this](https://github.com/AtomLinter/linter-pylint/blob/71463851dc3a3fec39c4cdebbfa6b9c09b25170f/lib/main.js#L67).

The reason I'm saying this should be done is you are currently depending on users being _exact_ in following your directions `foo/bar,foobar`. But as we all know users are the absolute worst at following directions and will do things like `foo/bar, foobar`, giving your current code a nice extra space in there. The array config type offloads all the extra massaging of the string to Atom and lets us just focus on using the values.

GreenGremlin: Are you saying that I can use project manager to set advancedLocalNodeModules on a per-project basis?

Arcanemagus: Exactly, `project-manager` lets you change settings based on the open project.

IanVS:
@GreenGremlin does `project-manager` meet your needs?

Arcanemagus:
Closing this since this can be managed by `project-manager`.

PR #868, opened 2017-03-31
Hiding 'ESLint: Fix file' in non-supported files

Xapphire13:


src/main.js
```diff
@@ -93,7 +94,16 @@ module.exports = {
       }
     }))
 
-    this.subscriptions.add(atom.commands.add('atom-text-editor', {
+    // Marks each JavaScript editor with a CSS class so that
+    // we can enable commands only for JavaScript editors.
+    this.subscriptions.add(atom.workspace.observeTextEditors((textEditor) => {
+      if (textEditor.getRootScopeDescriptor().scopes
```
Arcanemagus: Can you change this to match the [current checks](https://github.com/Xapphire13/linter-eslint/blob/bd2e81f176f81027a2b94a55a3310bf9b0a4ab97/src/main.js#L51), in fact that check should just be moved to a helper function and called in both places.

src/main.js
```diff
@@ -93,7 +94,16 @@ module.exports = {
       }
     }))
 
-    this.subscriptions.add(atom.commands.add('atom-text-editor', {
+    // Marks each JavaScript editor with a CSS class so that
+    // we can enable commands only for JavaScript editors.
```
Arcanemagus: This linter supports any file type the user desires since there are ESLint processors for some pretty strange templating languages out there. `JavaScript` should be changed to `supported`.

src/main.js
Xapphire13: The downside to the check you linked to is that it's only for the scope of where the current cursor is. Where as my code is checking the scope of the file itself.

Arcanemagus: That's actually the _upside_, not downside. Say for example you are editing an HTML file with some embedded JS code. The scope of the file would be `source.html`, while the scope of the embedded JS will be `source.js.embedded.html`. If we just went with your current method the file would never trigger linting on the embedded JS code, while checking the cursor means that if the user has a cursor in a scope that the linter provider understands it will be enabled.

This is how `linter` itself triggers providers in the first place, and one of the reasons I'm not exactly sold on the whole idea of adding a class to the editor instead of just checking things when called since we aren't really sure what situations the user can get into.

For other linters where there is a nice pre-defined list of possible scopes it might work, but for here I'm not sure I see how it will work without a ton of hacky workarounds (observing all editor cursors and checking if one ever enters a supported scope in the editor...).

Xapphire13: That makes sense and I agree. The issue I'm trying to address is when you have many linters installed, and you open the command pallet, start typing "fix" and then the fix commands for all your linters appear. Now you have the manual task of finding the right one.

The only way I know of how to scope commands is via CSS selectors (at least according to the [atom API](https://atom.io/docs/api/v1.15.0/CommandRegistry#instance-add)). With that in mind I see 2 options, either:

1.   Drop this PR and allow the command in all scopes
2.   Monitor each time the cursor moves, check the scope and add/remove the CSS selector
     - I don't like this option since it sounds pretty expensive and could slow the editor and ruin UX.

What are your feelings on this? Depending on your answser I will also amend my [TSLint PR](https://github.com/AtomLinter/linter-tslint/pull/149) accordingly.

Arcanemagus: Hmmmm, I think I would say that it will work for providers with well defined allowed scope lists (like TSLint), and in fact for those the `getRootScopeDescriptor` might even be a better choice.

For providers like this one though where the allowed scopes can change at any time, this doesn't really make much sense, and to make it work at all would involve watching quite a bit of the components of Atom to ensure you get everything, so unfortunately I think for `linter-eslint` at least this doesn't really make much sense.

Xapphire13: Okay, I shall abandon this PR and leave my TSLint PR as is. Thanks for your input =]

Arcanemagus: Thanks for sticking through with it while I figured out what felt "wrong" about it 😛.

I'll do a re-review of the TSLint one in a bit and most likely merge it.

Arcanemagus:
Closing based on the [above discussion](https://github.com/AtomLinter/linter-eslint/pull/868#discussion_r109778517).

PR #873, opened 2017-04-04
Use ESLint's CLIEngine API rather than the command line (fixes #797)

### LINKED ISSUE #797: Why does linter-eslint use the ESLint CLI executable instead of CLIEngine?
## Issue Type

Question

## Issue Description

This relates to @IanVS's comment at https://github.com/eslint/eslint/issues/7955#issuecomment-276217312:

> That said, we are using the cli executable directly (not using CLIEngine).

I'm interested in why linter-eslint does this. Naively, it seems like using CLIEngine to lint files would be more performant than using the executable, because running the executable requires waiting for Node to start up and for all the modules to be loaded. Is there some limitation in the CLIEngine or Atom API that prevents this from being a viable option?

not-an-aardvark:
This updates linter-eslint to use ESLint's [CLIEngine API](http://eslint.org/docs/developer-guide/nodejs-api#cliengine) rather than the private `cli` object. This makes the logic for interacting with ESLint much simpler.

I'm not very familiar with the linter-eslint codebase or Atom plugins in general, so be sure to review this thoroughly.


src/worker.js
```diff
@@ -54,16 +39,19 @@ create().onRequest('job', ({ contents, type, config, filePath, projectPath, rule
...
-    job.response = lintJob(argv, contents, eslint, configPath, config)
+  const noProjectConfig = (configPath === null || isConfigAtHomeRoot(configPath))
+  if (noProjectConfig && config.disableWhenNoEslintConfig) {
+    job.response = []
+  } else if (type === 'lint') {
+    const report = lintJob({ cliEngineOptions, contents, eslint, filePath })
+    job.response = report.results.length ? report.results[0].messages : []
   } else if (type === 'fix') {
-    job.response = fixJob(argv, eslint)
+    job.response = fixJob({ cliEngineOptions, contents, eslint, filePath })
```
IanVS: Question: Is the reason you're passing a currently-always-undefined `contents` into `fixJob` because you expect that in the future we would pass unsaved text into the fix job?  For some reason I thought that ESLint wouldn't allow `--fix` for stdin, but maybe it does when using `executeOnText`?  Could you talk a little about your thoughts here on passing `contents` to `fixJob`?

src/worker.js
```diff
@@ -10,41 +10,26 @@ import { isConfigAtHomeRoot } from './is-config-at-home-root'
...
 
-function fixJob(argv, eslint) {
-  const exit = eslint.execute(argv)
-  if (exit === 0) {
+function fixJob({ cliEngineOptions, contents, eslint, filePath }) {
+  const report = lintJob({ cliEngineOptions, contents, eslint, filePath })
+
+  eslint.CLIEngine.outputFixes(report)
+
+  if (!report.results.length || !report.results[0].messages.length) {
```
IanVS: Unfortunately, this won't work if a user tries to fix an ignored file.  We will get back a warning from ESLint that the file is ignored, and therefore the message displayed will be:

>Linter-ESLint: Fix attempt complete, but linting errors remain.

which is less than ideal.  I really wanted to be happy about deleting `ignoredMessages`, but I think we still need it, at least for this situation.  We could even be smart and show a meaningful message if we determine that the file the user tried to fix is being ignored.

src/worker.js
```diff
@@ -54,16 +39,19 @@ create().onRequest('job', ({ contents, type, config, filePath, projectPath, rule
...
 
-  const argv = Helpers.getArgv(type, config, rules, relativeFilePath, fileDir, configPath)
+  const cliEngineOptions =
+    Helpers.getCLIEngineOptions(type, config, rules, relativeFilePath, fileDir, configPath)
 
-  if (type === 'lint') {
-    job.response = lintJob(argv, contents, eslint, configPath, config)
+  const noProjectConfig = (configPath === null || isConfigAtHomeRoot(configPath))
+  if (noProjectConfig && config.disableWhenNoEslintConfig) {
+    job.response = []
```
IanVS: Good call moving this check down here, it should definitely apply to lint jobs and fix jobs. 👍

src/worker.js
```diff
@@ -54,16 +39,19 @@ create().onRequest('job', ({ contents, type, config, filePath, projectPath, rule
   const configPath = Helpers.getConfigPath(fileDir)
   const relativeFilePath = Helpers.getRelativePath(fileDir, filePath, config)
 
-  const argv = Helpers.getArgv(type, config, rules, relativeFilePath, fileDir, configPath)
+  const cliEngineOptions =
+    Helpers.getCLIEngineOptions(type, config, rules, relativeFilePath, fileDir, configPath)
```
IanVS: I'm not sure how @Arcanemagus feels about this one, but personally I would prefer avoiding the long-line error by formatting it as:

```js
const cliEngineOptions = Helpers.getCLIEngineOptions(
  type, config, rules, relativeFilePath, fileDir, configPath
)
```

src/worker.js
```diff
@@ -10,41 +10,26 @@ import { isConfigAtHomeRoot } from './is-config-at-home-root'
...
-  // V2.11.1
-  'File ignored because of a matching ignore pattern. Use "--no-ignore" to override.',
-  // supress warning that the current file is ignored by eslint by default
-  'File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern \'!<relative'
-    + '/path/to/filename>\'") to override.',
-  'File ignored by default. Use "--ignore-pattern \'!node_modules/*\'" to override.',
-  'File ignored by default. Use "--ignore-pattern \'!bower_components/*\'" to override.',
-]
+function lintJob({ cliEngineOptions, contents, eslint, filePath }) {
+  const cliEngine = new eslint.CLIEngine(cliEngineOptions)
```
IanVS: I wonder, are we going to pay a performance penalty here by creating a new `cliEngine` each time we perform a linting job?  If there's much overhead here, maybe it makes sense to memoize it, and only create a new `cliEngine` when the options change.  What do you think?

not-an-aardvark: That seems like it could be a good idea. It looks like  `CLIEngine` constructor checks the `--rulesdir` directory for updates only once when it's constructed, so this would mean that any changes to the rules in the rules directory wouldn't be reflected immediately. (I suspect it's pretty rare for users to modify the rules directory in the common case.)

Also, it's worth noting that the `cli` API that we previously used constructs a new instance of `CLIEngine` on every invocation too, so I doubt this will hurt performance compared to the previous behavior (although there might be room for improvement).

src/worker.js
not-an-aardvark: Actually, it's because I didn't realize `contents` was always `undefined`. 😄 

It's true that ESLint doesn't allow `--fix` from stdin on the command line, but this isn't a problem when calling `executeOnText` because it accepts a [`filename`](https://github.com/eslint/eslint/blob/f560c0660d9ab47c5592b0dd019404ba55f7b6ec/lib/cli-engine.js#L691) argument. So using `executeOnText` followed by `CLIEngine.outputFixes` would allow us to pass unsaved text (if we wanted to do that).

Notably, `executeOnText` also accepts a [warnIgnored](https://github.com/eslint/eslint/blob/f560c0660d9ab47c5592b0dd019404ba55f7b6ec/lib/cli-engine.js#L692) argument, which can be used to disable "file ignored because of matching ignore pattern" messages. If we switch to using `contents` and used `executeOnText` instead of `executeOnFiles`, we could also resolve the issue that you brought up in https://github.com/AtomLinter/linter-eslint/pull/873#discussion_r110074237 without needing to bring back the message list.

(I suppose another way to do this would be to read the file from the filesystem ourselves and call `executeOnText` on it to get rid of the ignore warnings.)

src/worker.js
Arcanemagus: That's how I would prefer it as well, but the styling in this project already has issues as far as I'm concerned so I just left it for a future update of the style 😛.

src/worker.js
IanVS: I think we should explore using `executeOnText` for fixes, but in a separate PR. I'd like to keep this PR small in scope so we're not changing too much at once.

not-an-aardvark: Ok, in that case should I add back the list of ignore messages for now?

src/worker.js
IanVS: Perhaps we can move to `executeOnText` for fix jobs at some point and turn off `warnIgnored`, as you mentioned in https://github.com/AtomLinter/linter-eslint/pull/873#discussion_r110076641, but for now let's keep the `ignoredMessages` and use them here.

src/worker.js
IanVS: That would be my preference, yes.

src/worker.js
IanVS: Cool, let's leave this how it is and possibly try to memoize it in another PR.

src/worker.js
Arcanemagus: Nevermind, just read the comments below, moving that to a separate PR works for me.

IanVS: Yes, that's the proposal.  I also think it would be easy, but I'd like it to be done in a separate PR to isolate the changes.

spec/linter-eslint-spec.js
```diff
@@ -162,13 +162,21 @@ describe('The eslint provider for Linter', () => {
...
         atom.workspace.open(ignoredPath).then(editor =>
           lint(editor).then(messages => expect(messages.length).toBe(0))
         )
       )
     })
+
+    it('will not give warnings when autofixing the file', () => {
+      waitsForPromise(() =>
+        atom.workspace.open(ignoredPath).then(editor =>
+          fix(editor).then(result => expect(result).toBe('Linter-ESLint: Fix complete.'))
```
IanVS: This works for now, and it was the previous behavior, but I think a better thing to do would be to actually tell the user that nothing happened because the file was ignored.  :)

Arcanemagus: It used to be that way, but users complained, which is why it's this way in the first place. Can't please everyone I guess.

Personally I would prefer to have the messages shown myself.

Arcanemagus:
Approved and merging, thanks again! I can't believe this ended up being such a simple change.

spec/linter-eslint-spec.js
IanVS: I think you mean we used to show the 'file ignored' messages when linting. And yeah, that's noisy. But when a user manually runs the fix command, we should show a meaningful message. I'll push a PR later if I can get around to it.

IanVS:
Agreed. Thanks @not-an-aardvark! We should have tried this a lot sooner. 😂

PR #875, opened 2017-04-06
Reduce package activation time

### LINKED ISSUE #871: Package activation time causes significant slowdown for new Atom windows
## Issue Type
Bug

## Issue Description
`linter-eslint` is doing too much synchronously during package activation:
* Requires should be switched to lazy requires
* Use of atom-package-deps should be put behind a debounce
* The worker should be spawned after activation
* Observation of text editor (and other) events should be moved out of activate and ideally behind a debounce

## Bug Checklist
- [x] Restart Atom
- [x] Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- [x] ~Paste the output of the `Linter Eslint: Debug` command from the Command Palette below~

screen shot 2017-04-03 at 11 16 08 am

Arcanemagus:
Reduce activation time of the package by deferring some actions for a later idle time and lazily requiring imports where possible. On my system this reduces package activation from an average of 177 ms down to 7 ms. The remaining 7 ms is largely taken up with Atom internals, with only 1.3 ms being under our control in any manner.

Fixes #871.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -9,7 +9,7 @@ import {
   generateDebugString,
 } from './helpers'
 import { getConfigPath } from './worker-helpers'
-import { isConfigAtHomeRoot } from './is-config-at-home-root'
+let isConfigAtHomeRoot
 
 // Configuration
 const scopes = []
@@ -67,6 +67,9 @@ module.exports = {
         if (validScope && atom.config.get('linter-eslint.fixOnSave')) {
           if (this.worker === null) {
             initializeWorker()
+          }
+          if (!isConfigAtHomeRoot) {
+            isConfigAtHomeRoot = require('./is-config-at-home-root')
           }
           const filePath = editor.getPath()
           const projectPath = atom.project.relativizePath(filePath)[0]
--- a/src/main.js
+++ b/src/main.js
@@ -4,6 +4,8 @@
 import { CompositeDisposable } from 'atom'
 
 // Dependencies
+// NOTE: We are not directly requiring these in order to reduce the time it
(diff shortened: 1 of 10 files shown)


Arcanemagus:
After rebasing on the fixed specs locally, some issues with the worker not being initialized quick enough have cropped up, looking into fixing this.

Arcanemagus:
Fixed 😉

src/main.js
```diff
@@ -47,17 +78,29 @@ module.exports = {
...
             scopes.includes(scope)))
         if (validScope && atom.config.get('linter-eslint.fixOnSave')) {
+          if (this.worker === null) {
+            await waitOnIdle()
+          }
+          if (!path) {
+            path = require('path')
+          }
+          if (!isConfigAtHomeRoot) {
+            isConfigAtHomeRoot = require('./is-config-at-home-root')
```
IanVS: Fix-on-save is currently broken, with an error message that `isConfigAtHomeRoot` isn't a function.  It seems that you need to use `isConfigAtHomeRoot = require('./is-config-at-home-root').default`, although to be honest, I'm not entirely clear on the "right" way to interop `require` statements with `export`.  Too bad we don't have `import()` yet.

Arcanemagus:
Whoops, that's the problem with having all of these branches running off of each other. Already found that and fixed it, but apparently that was in a different branch, and since the fixing of the specs that would expose that is over in #878 I missed it here. Will update in the morning!

Arcanemagus:
@IanVS `is-config-at-home-root` should be fixed now.

PR #876, opened 2017-04-06
Revert #853

Arcanemagus:
Reverts #853.

From the testing done so far this code either:

* Does exactly what Atom is already doing itself if the fix doesn't replace the entire file
* Doesn't actually work if the entire file is replaced (rare to reproduce this)

So in either case the code is doing nothing but adding complexity to the codebase.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -234,9 +234,6 @@ module.exports = {
       rules = ignoredRulesWhenFixing
     }
 
-    // The fix replaces the file content and the cursor jumps automatically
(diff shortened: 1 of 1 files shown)


Arcanemagus:
@yacut The GIF you posted falls under the first path outlined above, are you seeing any cases where the entire file is being replaced and the cursor is actually being restored?

Arcanemagus:
Note, tracking the fix for the likely issue that prompted this initially is here: https://github.com/atom/atom/issues/14189.

PR #878, opened 2017-04-08
Async-ify the specs

Arcanemagus:
Bring in the `jasmine-fix` module and utilize the `async` compatible `beforeEach` and `it` methods it provides. Refactor the code to utilize this instead of the `waitsForPromise` method from the default
environment.

Includes several fixes to the specs:
* Introduce a `getNotification` method to grab the next Notification added to Atom, allowing situations such as waiting on the fix command.
* Re-write the `makeFixes` function so it waits on the fix command to finish, including the file changes in Atom before continuing the spec.
* Make both `fix` specs use the same functions so the code isn't duplicated.
* Utilize the `lint` function directly, using the worker spawned from the `activate` call instead of a separate one spawned in the specs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "eslint": "^3.17.0",
     "eslint-config-airbnb-base": "^11.0.0",
     "eslint-plugin-import": "^2.2.0",
+    "jasmine-fix": "^1.0.1",
     "rimraf": "^2.5.4"
   },
   "package-deps": [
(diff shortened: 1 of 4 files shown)


IanVS:
>Utilize the lint function directly, using the worker spawned from the activate call instead of a separate one spawned in the specs.

😻

Arcanemagus:
Trying closing/reopening this to see if we can wake Travis-CI up...

PR #889, opened 2017-04-20
Atom Task API

Arcanemagus:
Moves us to utilizing Atom's Task API for managing the worker process.

This provides the framework for a future PR switching to using per-package transpilation, as well as reduces the external dependencies of the project.

It also appears that this provides a slight speedup to linting as average lints for a specific file went from 115 ms down to 104 ms!

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,38 +1,68 @@
 'use babel'
 
-import ChildProcess from 'child_process'
-import { createFromProcess } from 'process-communication'
 import { join } from 'path'
 import escapeHTML from 'escape-html'
 import ruleURI from 'eslint-rule-documentation'
 import { generateRange } from 'atom-linter'
+import cryptoRandomString from 'crypto-random-string'
 
 // eslint-disable-next-line import/no-extraneous-dependencies, import/extensions
-import { Disposable, Range } from 'atom'
+import { Range } from 'atom'
 
-export function spawnWorker() {
-  const env = Object.create(process.env)
-
-  delete env.NODE_PATH
-  delete env.NODE_ENV
-  delete env.OS
-
-  const child = ChildProcess.fork(join(__dirname, 'worker.js'), [], { env, silent: true })
-  const worker = createFromProcess(child)
-
-  child.stdout.on('data', (chunk) => {
-    console.log('[Linter-ESLint] STDOUT', chunk.toString())
-  })
-  child.stderr.on('data', (chunk) => {
-    console.log('[Linter-ESLint] STDERR', chunk.toString())
-  })
-
-  return {
-    worker,
-    subscription: new Disposable(() => {
-      worker.kill()
-    })
+/**
+ * Start the worker process if it hasn't already been started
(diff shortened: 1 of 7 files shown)


src/helpers.js
```diff
@@ -1,38 +1,68 @@
...
-  child.stderr.on('data', (chunk) => {
-    console.log('[Linter-ESLint] STDERR', chunk.toString())
-  })
+/**
+ * Send a job to the worker and return the results
+ * @param  {Task} worker The worker Task to use
+ * @param  {Object} config Configuration for the job to send to the worker
+ * @return {Object|String|Error}        The data returned from the worker
+ */
+export async function sendJob(worker, config) {
```
IanVS: 👍 for docblocks!

src/helpers.js
```diff
@@ -1,38 +1,68 @@
...
+  // Ensure the worker is started
+  try {
+    startWorker(worker)
+  } catch (e) {
+    throw e
```
IanVS: __*Question*__

What's the purpose of catching and re-throwing the error here (as well as in a few other places)?

src/helpers.js
```diff
@@ -1,38 +1,68 @@
 'use babel'
 
-import ChildProcess from 'child_process'
-import { createFromProcess } from 'process-communication'
 import { join } from 'path'
 import escapeHTML from 'escape-html'
 import ruleURI from 'eslint-rule-documentation'
 import { generateRange } from 'atom-linter'
+import cryptoRandomString from 'crypto-random-string'
```
IanVS: Just curious, why did you choose this package over, say, https://github.com/kelektiv/node-uuid?

src/main.js
```diff
@@ -216,14 +126,20 @@ module.exports = {
       ignoredRulesWhenFixing = idsToIgnoredRules(ids)
     }))
 
-    // Initialize the worker during an idle time, with a maximum wait of 5 seconds
-    window.requestIdleCallback(initializeWorker, { timeout: 5000 })
+    const initializeESLintWorker = () => {
+      this.worker = new Task(require.resolve('./worker.js'))
+    }
+    // Initialize the worker during an idle time
+    window.requestIdleCallback(initializeESLintWorker)
```
IanVS: I see you've removed the 5 second timeout.  Why's that?

src/helpers.js
Arcanemagus: We just need something "sufficiently random" to identify each run, a UUID would guarantee it... but that module is rather large in size, and takes 1 second (really!) to initialize, so I used this instead.

We could even just stringify some `Math.random()` output and probably be just fine here if you want to remove the module.

src/main.js
Arcanemagus: If you specify a timeout the idle callback can skip the queue entirely, so we would need to re-architecture how the entire thing works to ensure that the worker has initialized when needed.

src/helpers.js
Arcanemagus: I'm not entirely sure, will try removing them and see how it works 😉.

PR #890, opened 2017-04-20
Utilize Atom's built in transpilation

### LINKED ISSUE #857: Revamp transpilation pipeline
## Issue Type
Notes for enhancement

## Issue Description

[atom-babel6-transpiler](https://github.com/atom/atom-babel6-transpiler) allows us to tell Atom how to transpile the files automatically for us, and [babel-preset-env](https://github.com/babel/babel-preset-env) looks to be a decent place to start on a configuration.

Alternatively, here is an example Babel configuration provided by @binarymuse as an example:
```cjson
{
  "sourceMaps": "inline",
  "plugins": [
    "./graphql/relay-babel-plugin.js",
    "./assert-async-plugin.js",
    "./assert-messages-plugin.js",
    "transform-object-rest-spread",
    "transform-es2015-modules-commonjs",
    "transform-async-to-generator",
    "transform-decorators-legacy", // must come before class-properties
    "transform-class-properties",
    "transform-es2015-destructuring", // https://github.com/babel/babel/issues/4074
    "transform-es2015-parameters", // same
  ],
  "presets": [
    "react",
  ]
}
```

Arcanemagus:
Atom v1.13.0 introduced support for package's defining their own transpilation pipeline. Utilize this to remove the transpiled code from the repo and let Atom generate it on the fly.

**NOTE**: Based on the Task API branch, will need rebasing once https://github.com/AtomLinter/linter-eslint/pull/889 gets merged into `master`!

Fixes #857.


Arcanemagus:
Note that this has a detrimental effect on the "activation time" (as TimeCop calls it) of the package since the transpiler engine must be loaded:

image

That's ~29 ms added due to this.

Arcanemagus:
Filed https://github.com/atom/atom/issues/14300 about the overhead from utilizing this. I still feel this is the direction we should go as it massively simplifies the code base, and the overhead there is coming from Atom itself.

Arcanemagus:
Looks like Travis-CI and AppVeyor didn't like the base branch change, attempting a close/re-open.

Arcanemagus:
I think the AppVeyor failure is due to https://github.com/atom/ci/pull/66, will update to my test version of the script to check.

Arcanemagus:
Hmm, that fixes the usage of the system node for installing the dev modules, but it's still failing as if it can't install the `babel-preset-node5` module.

Arcanemagus:
AppVeyor seems to have had a temporary issue causing the build to fail for a reason that _should_ be completely out of their control, but changing nothing here has caused it to work properly.

On the plus side, I found that `.babelrc` was still in the project and have added removing that to this PR.

PR #898, opened 2017-05-01
Preserve `.eslintcache` on fixes

### LINKED ISSUE #635: `.eslintcache` is deleted when auto-fix is performed
ESLint since version `3.0.0` will no longer remove the `.eslintcache` file whenever a file is linted by linter-eslint.  (see https://github.com/AtomLinter/linter-eslint/issues/587 for discussion of that problem).

_However_, it will delete the cache when we auto-fix a file.  That's because we're not using `--stdin` when we perform fixing (see https://github.com/AtomLinter/linter-eslint/blob/master/src/worker-helpers.js#L109).

I think the solution here is to add an option like `Preserve ESLint cache`, which would add `--cache` when performing a fix.

IanVS:
Fixes #635 

The main point of this PR is to avoid deleting the `.eslintcache` file when performing fix jobs in `linter-eslint`.  Previously, they were deleted because if ESLint is run without the `cache` option, it will automatically delete any existing cache files.  `linter-eslint` doesn't delete the cache on regular `lint` jobs because it uses `executeOnText`, which bypasses the cache and does not delete it.  

**Update**: I've changed this PR to use `executeOnText` for fix jobs as suggested by @not-an-aardvark in https://github.com/AtomLinter/linter-eslint/pull/873#discussion_r110076641.  It's a much better solution than my first approach.

~~The approach I took was fairly simplistic.  We will search for a `.eslintcache` file, and if we find it, we will set `cache` for fix jobs.  There is a slight danger here that we may create extra `.eslintcache` files in unexpected locations, because that setting will also _create_ cache files if none exist.  In my own testing so far I haven't experienced that problem, but we should be on the lookout for it in more complicated project structures.~~

~~At this time I think it's reasonable to only consider files named `.eslintcache`, as that is the default and most common.  If there is overwhelming demand for a config setting, we can consider that, but for now I'd like to keep it as simple as possible.~~

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -15,6 +15,7 @@ const badPath = path.join(fixturesDir, 'files', 'bad.js')
 const badInlinePath = path.join(fixturesDir, 'files', 'badInline.js')
 const emptyPath = path.join(fixturesDir, 'files', 'empty.js')
 const fixPath = path.join(fixturesDir, 'files', 'fix.js')
+const cachePath = path.join(fixturesDir, 'files', '.eslintcache')
 const configPath = path.join(fixturesDir, 'configs', '.eslintrc.yml')
 const importingpath = path.join(fixturesDir,
   'import-resolution', 'nested', 'importing.js')
@@ -43,6 +44,12 @@ function copyFileToDir(fileToCopyPath, destinationDir) {
   })
 }
 
+/**
+ * Utility helper to copy a file into the OS temp directory.
+ *
+ * @param  {string} fileToCopyPath  Path of the file to be copied
+ * @return {string}                 Full path of the file in copy destination
+ */
 function copyFileToTempDir(fileToCopyPath) {
   return new Promise(async (resolve) => {
     const tempFixtureDir = fs.mkdtempSync(tmpdir() + path.sep)
@@ -60,6 +67,25 @@ async function getNotification() {
     }
     // Subscribe to Atom's notifications
     notificationSub = atom.notifications.onDidAddNotification(newNotification)
+  })
+}
+
+async function makeFixes(textEditor) {
+  return new Promise(async (resolve) => {
+    // Subscribe to the file reload event
+    const editorReloadSub = textEditor.getBuffer().onDidReload(async () => {
+      editorReloadSub.dispose()
+      // File has been reloaded in Atom, notification checking will happen
+      // async either way, but should already be finished at this point
+      resolve()
+    })
+
+    // Now that all the required subscriptions are active, send off a fix request
+    atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
+    const notification = await getNotification()
+
+    expect(notification.getMessage()).toBe('Linter-ESLint: Fix complete.')
+    expect(notification.getType()).toBe('success')
   })
 }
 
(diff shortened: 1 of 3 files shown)


IanVS:
Now that I think about this a little more, we should probably explore using `executeOnText` for fixes as suggested in https://github.com/AtomLinter/linter-eslint/pull/873#discussion_r110076641.  That would be a more robust solution than what I'm doing here, and is worth a shot.

Also, I guess the Appveyor spec failure is the same race condition failure that caused us to disable the other test in that suite.  Maybe breaking them into separate describe blocks will help… I'll try it later.

Arcanemagus:
That's a bizarre failure in the AppVeyor build, it looks like it grabbed the notification for one of the debug tooltip tests, I'm not entirely sure we can actually fix that within the limitations of Jasmine 1.3 that Atom uses 😕.

src/main.js
```diff
@@ -162,12 +167,15 @@ module.exports = {
...
+        if (text.length === 0) {
+          return
+        }
         const filePath = textEditor.getPath()
         const projectPath = atom.project.relativizePath(filePath)[0]
 
-        if (!textEditor || textEditor.isModified()) {
+        if (!textEditor) {
           // Abort for invalid or unsaved text editors
-          atom.notifications.addError('Linter-ESLint: Please save before fixing')
```
IanVS: Note, we can now perform fixes on unsaved text!

spec/linter-eslint-spec.js
```diff
@@ -28,20 +29,28 @@ const modifiedIgnoreSpacePath = path.join(fixturesDir,
...
-      })
-    )
+    const destinationPath = path.join(destinationDir, path.basename(fileToCopyPath))
+    const ws = fs.createWriteStream(destinationPath)
+    ws.on('close', () => resolve(destinationPath))
     fs.createReadStream(fileToCopyPath).pipe(ws)
   })
 }
 
+function copyFileToTempDir(fileToCopyPath) {
```
Arcanemagus: docbloc this?

src/worker.js
```diff
@@ -36,8 +36,8 @@ function lintJob({ cliEngineOptions, contents, eslint, filePath }) {
     : cliEngine.executeOnFiles([filePath])
```
Arcanemagus: Closest I can get to commenting while still be in the PR:

`contents` will _always_ be a string now, so this entire check is irrelevant now.

IanVS:
@Arcanemagus thanks for the review.  I've addressed your feedback and squashed commits.

src/main.js
IanVS: There is a small caveat here.  If you make a change to introduce a linting error within a previously-passing file without saving, and the fix results in the same code as what was previously saved, manually running a fix command will have no visible effect in the editor.  That's because the file on disk will essentially not be changed, and so Atom won't know to reload the editor.  I think that's a small enough edge case that we don't need to worry about it, but wanted to make you aware.

Arcanemagus:
Currently although we can get the list of fixes for an unsaved buffer, since we are using [`CLIEngine.outputFixes()`](http://eslint.org/docs/developer-guide/nodejs-api#outputfixes) to _apply_ them, we must still have the buffer be saved when running or it appears nothing happens. (Or worse, it tries to apply fixes from the unsaved buffer to the saved state on disk, with potentially disastrous results.)

IanVS:
Hah!  I spend a few minutes in your non-compiling branch, and now I'm already forgetting to check in compiled files!

PR #902, opened 2017-05-03
Wait on specific Notification in specs

Arcanemagus:
As the specs execute asynchronously, it's possible for there to be more than one Notification observer active at a time, leading to some specs intermittently failing when they grab the "wrong" notification. Enhance the function that waits on a notification to check for an expected message as well, and silently discard the received notification if it doesn't match what the spec is looking for.


Arcanemagus:
Merging without review as this fixes inconsistencies in specs on the current state of `master`, and it's a rather small change.

Issue #780, opened 2017-01-11
[Help]Can not find global eslint installed with yarn


linkdesu:
Yarn install eslint in `~/.config/yarn/global/node_modules/.bin/eslint` and create a soft link at `/usr/local/Cellar/node/7.2.1/bin/eslint`. My $PATH contains `/usr/local/Cellar/node/7.2.1/bin`, so I'm sure can execute eslint from anywhere and I've checked the config `Use global ESLint installation`. 

But unfortunately, linter-eslint keep alert `Error: ESLint not found, Please install or make sure Atom is getting $PATH correctly`.

Did I still miss something important?

IanVS:
Can you please provide the output of the debug command as requested in the
issue template?

On Tue, Jan 10, 2017, 23:26 Aolin Xie <notifications@github.com> wrote:

Yarn install eslint in ~/.config/yarn/global/node_modules/.bin/eslint and
create a soft link at /usr/local/Cellar/node/7.2.1/bin/eslint. My $PATH
contains /usr/local/Cellar/node/7.2.1/bin, so I'm sure can execute eslint
from anywhere and I've checked the config Use global ESLint installation.

But unfortunately, linter-eslint keep alert Error: ESLint not found, Please
install or make sure Atom is getting $PATH correctly.

Did I still miss something important?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/AtomLinter/linter-eslint/issues/780>, or mute the thread

linkdesu:
Which debug command? Do you mean this:

```
Error: ESLint not found, Please install or make sure Atom is getting $PATH correctly
    at findESLintDirectory (/Users/linkdesu/.atom/packages/linter-eslint/lib/worker-helpers.js:90:13)
    at getESLintFromDirectory (/Users/linkdesu/.atom/packages/linter-eslint/lib/worker-helpers.js:100:30)
    at Object.getESLintInstance (/Users/linkdesu/.atom/packages/linter-eslint/lib/worker-helpers.js:127:10)
    at /Users/linkdesu/.atom/packages/linter-eslint/lib/worker.js:68:24
    at Emitter.emit (/Users/linkdesu/.atom/packages/linter-eslint/node_modules/sb-event-kit/lib/emitter.js:70:19)
    at /Users/linkdesu/.atom/packages/linter-eslint/node_modules/sb-communication/lib/main.js:22:23
[...]

PR #906, opened 2017-05-10
Fix for Yarn installations

bennypowers:
might break non-yarn non-win32 installations?


Arcanemagus:
As stated in https://github.com/AtomLinter/linter-eslint/issues/780#issuecomment-300612700, this reduces down to the same path as the `process.platform === 'win32'` path.

I'm going to close this as I'll be putting up a PR fixing this properly, as well as a few other minor issues related to this area shortly.

PR #907, opened 2017-05-10
Support global Yarn installs and better errors

### LINKED ISSUE #680: Error message is totally misleading - Error: ESLint not found, Please install or make sure Atom is getting $PATH correctly

### LINKED ISSUE #780: [Help]Can not find global eslint installed with yarn

Arcanemagus:
Add support for global Yarn installations. Apparently one of the few things Yarn did right is that all platforms use the same directory structure under the base directory, unfortunately that meant that on non-Windows platforms specifying the path still got you an unusable setup.

Also fixes the error messages that can be thrown to better reflect the actual causes of the messages.

Fixes #680.
Fixes #780.

### DIFF
--- a/spec/fixtures/global-eslint/node_modules/eslint/lib/api.js
+++ b/spec/fixtures/global-eslint/node_modules/eslint/lib/api.js
@@ -1,1 +1,0 @@
-module.exports = "located"
--- a/spec/fixtures/global-eslint/node_modules/eslint/package.json
+++ b/spec/fixtures/global-eslint/node_modules/eslint/package.json
@@ -1,3 +1,0 @@
-{
-  "main": "lib/api.js"
-}
--- a/spec/worker-helpers-spec.js
+++ b/spec/worker-helpers-spec.js
@@ -5,6 +5,11 @@ import * as Helpers from '../src/worker-helpers'
 
 const getFixturesPath = path => Path.join(__dirname, 'fixtures', path)
 
+
+const globalNodePath = process.platform === 'win32' ?
+  Path.join(getFixturesPath('global-eslint'), 'lib') :
+  getFixturesPath('global-eslint')
+
 describe('Worker Helpers', () => {
   describe('findESLintDirectory', () => {
     it('returns an object with path and type keys', () => {
@@ -25,7 +30,6 @@ describe('Worker Helpers', () => {
 
     it('does not find a local eslint when useGlobalEslint is true', () => {
       const modulesDir = Path.join(getFixturesPath('local-eslint'), 'node_modules')
-      const globalNodePath = getFixturesPath('global-eslint')
       const config = { useGlobalEslint: true, globalNodePath }
       const foundEslint = Helpers.findESLintDirectory(modulesDir, config)
       const expectedEslintPath = Path.join(getFixturesPath('local-eslint'), 'node_modules', 'eslint')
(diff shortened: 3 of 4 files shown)


Arcanemagus:
Hmmm, the spec failure is due to the fact that all platforms now check the `node_modules/eslint` path first, so it's finding that path even on the non-Windows platforms.

Do we want to split that folder up, or just remove the `lib` variant?

Arcanemagus:
Released in v8.2.0 🎉.

selfagency:
I have 8.2.1, I have eslint installed via yarn, and I cannot get it to find eslint even though it's in the $PATH and works fine on the command line. My yarn global prefix is /usr/local and the symlink to the yarn node_modules folder is in the /usr/local/bin folder.

Arcanemagus:
@selfagency The _only_ thing that this package uses `$PATH` for is resolving `npm` when attempting to get the prefix automatically. Since you are using `yarn` you will need to specify it manually. Please see #780 for some locations where it might be.

Note that using a global ESLint install is very much discouraged, but _supported_, you should be using a local instance in your project instead!

If you are still having issues please file a new issue.

PR #911, opened 2017-05-18
Handle fatal syntax errors gracefully

steelbrain:
ESLint outputs context in the most recent versions which is useless inside Atom

Can be seen in a lot of issues across the repos, https://github.com/steelbrain/linter-ui-default/issues/309 is the most recent one

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -208,8 +208,9 @@ const generateInvalidTrace = async (
  */
 export async function processESLintMessages(response, textEditor, showRule, worker) {
   return Promise.all(response.map(async ({
-    message, line, severity, ruleId, column, fix, endLine, endColumn
+    fatal, message: originalMessage, line, severity, ruleId, column, fix, endLine, endColumn
   }) => {
+    const message = fatal ? originalMessage.split('\n')[0] : originalMessage
     const filePath = textEditor.getPath()
     const textBuffer = textEditor.getBuffer()
     let linterFix = null


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Upcoming
```
Arcanemagus: Changelog here is generated automatically during release, so this should be removed.

spec/worker-helpers-spec.js
```diff
@@ -1,7 +1,7 @@
 'use babel'
 
 import * as Path from 'path'
-import * as Helpers from '../src/worker-helpers'
+import * as Helpers from '../src/worker/helpers'
```
Arcanemagus: There is no need to rearrange the files here.

src/helpers.js
```diff
@@ -208,8 +208,9 @@ const generateInvalidTrace = async (
  */
 export async function processESLintMessages(response, textEditor, showRule, worker) {
   return Promise.all(response.map(async ({
-    message, line, severity, ruleId, column, fix, endLine, endColumn
+    fatal, message: originalMessage, line, severity, ruleId, column, fix, endLine, endColumn
   }) => {
+    const message = fatal ? originalMessage.split('\n').shift() : originalMessage
```
Arcanemagus: Since the Array being created here is being discarded within the same command, just do `[0]` on the end, it's clearer what the intent is.

PR #915, opened 2017-05-19
Utilize warnIgnored instead of filtering messages

### LINKED ISSUE #900: Utilize warnIgnored
## Issue Type
Enhancement

## Issue Description
As brought up in https://github.com/AtomLinter/linter-eslint/pull/873#discussion_r110076641, now that we are using the CLIEngine we can just pass `warnIgnored` instead of parsing the output for "file ignored" messages.

Arcanemagus:
Now that all execution of ESLint is being done through `executeOnText` we can move to the `warnIgnored` option instead of manually filtering out the ignore messages.

Fixes #900.

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -142,6 +142,7 @@ export function getCLIEngineOptions(type, config, rules, filePath, fileDir, give
   const cliEngineConfig = {
     rules,
     ignore: !config.disableEslintIgnore,
+    warnIgnored: false,
     fix: type === 'fix'
   }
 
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Turns out this was _entirely_ wrong. The option has no meaning there and needed to be passed in as a third argument to `executeOnText`, not as an argument to `CLIEngine`.

The reason we never saw that this wasn't working is that the default behavior was fixed to be what you would expect anyway in https://github.com/eslint/eslint/pull/6632.

PR #958, opened 2017-07-10
fix: handle null projectPath

### LINKED ISSUE #854: linter-eslint not run on file with extension .js.erb
## Issue Type
Bug

## Issue Description
When I open file with extension .js.erb, eslint doesn't run on it.

So in List of scopes to run ESLint on. I added ', source.js.erb' to default string. However it stills not run.

leggsimon:
Fixes #854

### DIFF
--- a/spec/worker-helpers-spec.js
+++ b/spec/worker-helpers-spec.js
@@ -56,6 +56,22 @@ describe('Worker Helpers', () => {
       expect(foundEslint.path).toEqual(expectedBundledPath)
       expect(foundEslint.type).toEqual('bundled fallback')
     })
+
+    it('should handle a null projectPath when resolving path', () => {
+      const modulesDir = Path.join(getFixturesPath('local-eslint'), 'node_modules')
+      const config = {
+        advancedLocalNodeModules: '../node_modules',
+        useGlobalEslint: false
+      }
+
+      const findEslint = () => Helpers.findESLintDirectory(modulesDir, config, null)
+      expect(findEslint).not.toThrow()
+
+      const foundEslint = findEslint()
+      const expectedBundledPath = Path.join(__dirname, config.advancedLocalNodeModules, 'eslint')
+      expect(foundEslint.path).toEqual(expectedBundledPath)
+      expect(foundEslint.type).toEqual('advanced specified')
+    })
   })
 
   describe('getESLintInstance && getESLintFromDirectory', () => {
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -60,7 +60,7 @@ export function findESLintDirectory(modulesDir, config, projectPath) {
     eslintDir = Path.join(config.advancedLocalNodeModules || '', 'eslint')
   } else {
     locationType = 'advanced specified'
-    eslintDir = Path.join(projectPath, config.advancedLocalNodeModules, 'eslint')
+    eslintDir = Path.join(projectPath || __dirname, config.advancedLocalNodeModules, 'eslint')
   }
   if (isDirectory(eslintDir)) {
     return {


src/worker-helpers.js
```diff
@@ -60,7 +60,7 @@ export function findESLintDirectory(modulesDir, config, projectPath) {
     eslintDir = Path.join(config.advancedLocalNodeModules || '', 'eslint')
   } else {
     locationType = 'advanced specified'
-    eslintDir = Path.join(projectPath, config.advancedLocalNodeModules, 'eslint')
+    eslintDir = Path.join(projectPath || __dirname, config.advancedLocalNodeModules, 'eslint')
```
Arcanemagus: This most definitely isn't what you want here, if `null` is really getting legitimately passed in here, it should be handled differently.

spec/worker-helpers-spec.js
```diff
@@ -56,6 +56,22 @@ describe('Worker Helpers', () => {
...
     })
+
+    it('should handle a null projectPath when resolving path', () => {
+      const modulesDir = Path.join(getFixturesPath('local-eslint'), 'node_modules')
+      const config = {
+        advancedLocalNodeModules: '../node_modules',
+        useGlobalEslint: false
+      }
+
+      const findEslint = () => Helpers.findESLintDirectory(modulesDir, config, null)
```
Arcanemagus: Explicitly calling it with `null` here supposedly replicates what you are seeing, but I'd rather this test replicate the actual situation that triggers this case, not forcing it like this.

The tests should always replicate the real world case where possible.

leggsimon:
Firstly thanks a lot for reviewing and for the comments 👍 

> "can you describe _how_ this is even being triggered for you?"

No. I currently don't know enough about the internals of Atom and what [`jobConfig`](https://github.com/AtomLinter/linter-eslint/blob/master/src/worker.js#L30) is or why `projectPath` might have been `null`.

Although I may be very wrong, it seems to me as though `projectPath` must be available as some sort of global property that Atom provides. Because although it wasn't being passed in [here](https://github.com/AtomLinter/linter-eslint/blob/master/src/worker.js#L58) when I was logging it in the `findESLintDirectory` function it was `null` rather than `undefined` which is what I would've expected. This was the reason I was explicitly passing it in in the test.

>Normal linting of code [forces](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/worker-helpers.js#L104) it to be `''` if it was `null`.

Yep, I agree that is probably a better way to handle that, so I will change that.


I totally agree that "tests should always replicate the real world case" and short of testing the [`worker.js`](https://github.com/AtomLinter/linter-eslint/blob/master/src/worker.js#L28-L62) and having an integration test of sorts, I would argue I am testing the real world scenario of `projectPath` being a null value within the `findESLintDirectory` function.

I'm totally happy to have another crack at this at some point this week. It'd be great if you were able to give a short explanation of when `worker.js` is called and what the `process.on('message', () => {...})` is for/does so that I can try and replicate?

Arcanemagus:
I'm about 99% sure that what @IanVS pointed out is the real problem here.

`findESLintDirectory` is only called in 2 places: [here](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/worker-helpers.js#L80) and [here](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/worker.js#L58).

The first call point is only called from `getESLintInstance`, which forces it to be `''` if it's "falsy". The second call point, the one pointed out by @IanVS, doesn't even pass in the parameter so I'm pretty sure that's where your issue is getting triggered.

`debug` jobs are triggered by the [`getDebugInfo`](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/helpers.js#L87) function, which is triggerd (eventually) either manually through the `linter-eslint:debug` command, or the likely trigger for you hitting this: Generating the debug text for when an invalid point is encountered from ESLint.

tl;dr: I'm guessing you either are using an old version of ESLint that has a bug, or managed to find a new bug in ESLint. When gathering the debug information to generate an issue here it hit the bug in `linter-eslint` that lead to your issue. Passing in the `projectPath` [here]([here](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/worker.js#L58)), with the same forcing to `''` as the other place just in case, should fix this.

----

> I am testing the real world scenario of projectPath being a null value within the `findESLintDirectory` function.

If you can't answer _how_ that is getting called with `null`, then you aren't testing the real situation of where it is getting called with `null` 😛.

> It'd be great if you were able to give a short explanation of when `worker.js` is called and what the `process.on('message', () => {...})` is for/does so that I can try and replicate?

`worker.js` is a script that runs in a separate process from the main Atom renderer, so running ESLint doesn't slow down the editor when you are using it. It is initially started [here](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/main.js#L136). "Jobs" are sent to it using the [`sendJob`](https://github.com/AtomLinter/linter-eslint/blob/v8.2.1/src/helpers.js#L33) function which sends a message to the worker process and waits on the response. The `process.on('message'` is how the worker listens for these messages in order to know what it needs to do.

leggsimon:
Great, thanks for that, will work on this at some point this week and update the PR

PR #963, opened 2017-07-22
Add an ESLint Fix right click command

### LINKED ISSUE #955: Context menu for fixing current file
## Issue Type
Feature Request

## Issue Description

This was previously discussed in https://github.com/AtomLinter/linter-eslint/issues/940 but as a support request. The explanation (use the command palette), while accurate isn't user friendly. Fixing files on-demand is a common use case, and something supported by other editors like WebStorm.

In atom, the hacky way to do it is to edit the Init Script

```cs
atom.contextMenu.add {
  'atom-text-editor, .overlayer': [
    { type: 'separator' },
    {
      label: 'ESLint Fix',
      command: 'linter-eslint:fix-file'
    }
  ]
}
```

screen shot 2017-06-30 at 1 51 51 pm

I don't know enough about `shouldDisplay` and Atom to make it only show for JavaScript files, but that's the API to use.

Arcanemagus:
Add an "ESLint Fix" command to the right click menu when it is the active editor and is listed as a valid scope according to the current settings.

Fixes #955.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -41,6 +41,10 @@ const waitOnIdle = async () =>
     idleCallbacks.add(callbackID)
   })
 
+const validScope = editor => editor.getCursors().some(cursor =>
+  cursor.getScopeDescriptor().getScopesArray().some(scope =>
+    scopes.includes(scope)))
+
 module.exports = {
   activate() {
     let callbackID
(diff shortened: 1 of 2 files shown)


Arcanemagus:
@IanVS Do we want to take on this "balck magic", or is there a better way?

src/main.js
```diff
@@ -132,6 +133,26 @@ module.exports = {
...
+        shouldDisplay: (evt) => {
+          const activeEditor = atom.workspace.getActiveTextEditor()
+          if (!activeEditor) {
+            return false
+          }
+          // Black magic!
+          // Compares the private component property of the active TextEditor
+          //   against the components of the elements
+          const evtIsActiveEditor = evt.path.some(elem =>
+            elem.component && elem.component === activeEditor.component)
```
IanVS: Note: it seems that `activeEditor` does not have a `component` prop in atom `1.18.0`, so this only works in `1.19.0` and up.  I think that's fine, personally.

Arcanemagus: Fixed for Atom v1.18.0 😉.

fluxsauce:
Awesome, thank you!

PR #969, opened 2017-07-24
Handle entire file messages

### LINKED ISSUE #908: Missing rules result in a large report range
## Issue Type

Bug, sort of

## Issue Description

When using ESLint 4.0.0-alpha, I added a rule that does not exist to my config file. As expected, ESLint reported an error because the rule was missing. However, linter-eslint displayed the error like this:

In other words, it highlighted the entire file rather than a single line, which is what it did previously.

The issue is caused by [this change](http://eslint.org/docs/4.0.0/user-guide/migrating-to-4.0.0#report-locations) in ESLint 4.0. If a rule reports a node, the reported problem now has an end range. In this case, the entire AST was reported, so there are a lot of highlighted lines.

It might be worth limiting the number of lines highlighted in cases like this.

## Bug Checklist
- Restart Atom
- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not

```
Atom version: 1.16.0
linter-eslint version: 8.1.7
ESLint version: 4.0.0-alpha.1
Hours since last Atom restart: 0.1
Platform: darwin
Using local project ESLint from: /path/to/eslint/node_modules/eslint
Current file's scopes: [
  "source.js",
  "meta.function.js",
  "meta.parameters.js",
  "punctuation.definition.parameters.begin.bracket.round.js"
]
  "disableFSCache": true,
  "eslintRulesDir": "lib/internal-rules",
  "rulesToSilenceWhileTyping": [
    "no-trailing-spaces"
  ],
  "lintHtmlFiles": false,
  "useGlobalEslint": false,
  "showRuleIdInMessage": true,
  "eslintrcPath": "",
  "globalNodePath": "",
  "disableEslintIgnore": false,
  "fixOnSave": false,
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic"
  ],
  "rulesToDisableWhileFixing": []
}
```

Arcanemagus:
ESLint v4 can mark the root node of a file as having an error, which would lead to the entire file being reported as the range for the file.

Check for these messages and force them to instead highlight just the first line of the file.

Fixes #908.

### DIFF
--- a/spec/fixtures/configs/invalid-rule/.eslintrc.js
+++ b/spec/fixtures/configs/invalid-rule/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  root: true,
+  rules: {
+    'foo-bar': 'error'
+  }
+}
--- a/spec/fixtures/configs/invalid-rule/foo.js
+++ b/spec/fixtures/configs/invalid-rule/foo.js
@@ -1,0 +1,4 @@
+var foo = 42;
+
+// Add more code that will get highlighted by a root node message.
+var bar;
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -28,6 +28,7 @@ const modifiedIgnoreSpacePath = path.join(fixturesDir,
   'modified-ignore-rule', 'foo-space.js')
 const endRangePath = path.join(fixturesDir, 'end-range', 'no-unreachable.js')
 const badCachePath = path.join(fixturesDir, 'badCache')
+const invalidRulePath = path.join(fixturesDir, 'configs', 'invalid-rule', 'foo.js')
 
 /**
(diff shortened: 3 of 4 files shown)


Arcanemagus:
Pending [further discussion](https://github.com/AtomLinter/linter-eslint/issues/908#issuecomment-317540837) in #908.

Arcanemagus:
Closing this, it turns out this would work for the very simple case I was testing, but fixing it correctly is far more involved:

The range highlighted is the entire _code_ block, meaning leading and trailing whitespace _and comments_ are ignored. To fix this properly here we would need to be able to figure out the entire code range ourselves which seems like far more complexity and trouble than it's worth.

If we decide to revisit this later, this can serve as a starting point.

PR #978, opened 2017-08-09
Ignore invalid TextEditors for fix jobs

### LINKED ISSUE #972: Please save before fixing
## Issue Type
Bug

## Issue Description
I occasionally see the message "Please save before fixing", without any further explanation.
This happens usually when I do a replace in file in multiple files across a project (with one or more of those files open in Atom).

Why does this message appear? What can be done about it? This didn't use to appear before a few versions back.

## Bug Checklist
- Restart Atom
Yes, also happens after restart.

- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
Not applicable here.

```
Atom version: 1.20.0-dev-19b1810
linter-eslint version: 8.2.1
ESLint version: 4.3.0
Hours since last Atom restart: 0
Platform: darwin
Using global ESLint from: /usr/local/lib/node_modules/eslint
Current file's scopes: [
  "source.js",
  "punctuation.terminator.statement.js"
]
  "fixOnSave": true,
  "useGlobalEslint": true,
  "lintHtmlFiles": false,
  "showRuleIdInMessage": true,
  "eslintrcPath": "",
  "globalNodePath": "",
  "eslintRulesDir": "",
  "disableEslintIgnore": false,
  "disableFSCache": false,
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic"
  ],
  "rulesToSilenceWhileTyping": [],
  "rulesToDisableWhileFixing": []
}
```

Arcanemagus:
When a fix job is triggered but the TextEditor is invalid by the time the job runs silently return.

Fixes #972.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -227,7 +227,12 @@ module.exports = {
   async fixJob(isSave = false) {
     const textEditor = atom.workspace.getActiveTextEditor()
 
-    if (!textEditor || textEditor.isModified()) {
+    if (!textEditor || !atom.workspace.isTextEditor(textEditor)) {
+      // Silently return if the TextEditor is invalid
+      return
+    }
+
+    if (textEditor.isModified()) {
       // Abort for invalid or unsaved text editors
       const message = 'Linter-ESLint: Please save before fixing'
       atom.notifications.addError(message)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1964 @@
+{
+  "name": "linter-eslint",
+  "version": "8.2.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "acorn": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.1.1.tgz",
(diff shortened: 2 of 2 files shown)


Arcanemagus:
Blocked on https://github.com/AtomLinter/linter-eslint/pull/979.

src/main.js
```diff
@@ -227,7 +227,12 @@ module.exports = {
   async fixJob(isSave = false) {
     const textEditor = atom.workspace.getActiveTextEditor()
 
-    if (!textEditor || textEditor.isModified()) {
+    if (!textEditor || !atom.workspace.isTextEditor(textEditor)) {
+      // Silently return if the TextEditor is invalid
+      return
```
IanVS: We talked about maybe logging something out to the console, did you decide not to?

Arcanemagus: The issue I see with that is the initially reported scenario: Making changes to dozens (if not hundreds) of files at once. If that's what was happening your console would suddenly be spammed with messages with no way to determine what went wrong, or how to fix it (which you couldn't anyway here).

PR #981, opened 2017-08-09
Disable spying on Date.now by Jasmine

Arcanemagus:
It turns out that `TextBuffer::onDidReload` is being broken _only_ in the specs, as it relies on `Date.now()` which Jasmine was mocking. We can fix this by telling Jasmine to stop spying on that function.

Also re-enables a long-disabled test that looks to be working properly now that the `TextBuffer` rewrite has landed.

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -109,6 +109,9 @@ describe('The eslint provider for Linter', () => {
     await atom.packages.activatePackage('language-javascript')
     // Activate the provider
     await atom.packages.activatePackage('linter-eslint')
+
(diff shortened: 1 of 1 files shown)


Arcanemagus:
For reference, this was diagnosed in https://github.com/atom/atom/issues/15200 by @maxbrunsfeld.

Arcanemagus:
Hmmm, this seems to be causing random failures on AppVeyor only.

Arcanemagus:
Ah ha, duh, https://github.com/steelbrain/jasmine-fix/issues/6 was fixed, meaning `jasmine-fix` is doing this for us and now our attempt to do it is causing problems.

PR #1008, opened 2017-09-09
remove superfluous async & Promise from some specs

skylize:
Removes some uses of -new Promise- and -async/await- that do not serve any real purpose because underlying function calls already return promise


skylize:
Sorry, trying out `hub` for cli pull requests and submitted this wrong.

Issue #991, opened 2017-08-26
`eslintIgnore` property set in `package.json` is ignored


skylize:
## Issue Type
Bug

## Issue Description
`eslintIgnore` property set in `package.json` is ignored by Atom linter. 

## Details
According to [ESLint docs](https://eslint.org/docs/user-guide/configuring), "If an .eslintignore file is not found and an alternate file is not specified, eslint will look in package.json for an eslintIgnore key to check for files to ignore."

## Basic test

Files to demonstrate bug:

`build/invalid.js`
```javascript
An arbitrary raw string is not valid javascript.
```
`package.json`
``` javascript
{
// other package.json properties
//...
"eslintIgnore": "build/*"
}
```

**Expected behavior:** 
Open `build/invalid.js` in Atom and see no linting errors reported.

**Actual behavior:**
Open `build/invalid.js` in Atom and linter reports:
`Severity: Error`|`Provider: ESLint`|`Description: Parsing error: Unexpected token raw (Fatal)`|`Line: 1:3`

## Related tests
1. _Does eslint recognize this property if not running through Atom?_ 
```bash
path/to/project$ node_modules/.bin/eslint build/*

path/to/project/build/invalid.js
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override
```
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; _Yes_. 

Eslint binary picks up the rule from `package.json` and ignores the file as promised.

---
2. _Does it work if we use a `.eslintignore` file instead?_

`.eslintignore`
```text
build/*
```
Open `bu
`build/invalid.js` in Atom. No errors reported.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;_Yes_.

Atom plugin is not having any issue with applying the ignore rules, it just isn't picking up the rules from `package.json`.

Arcanemagus:
For reference this was added in https://github.com/eslint/eslint/pull/8690 which seems to have first appeared in ESLInt v4.1.0.

IanVS:
Thanks for the report, @skylize.  Would you be willing to submit a PR to fix this?

skylize:
@IanVS. If you can point me in the right direction, I'd be happy to at give it a shot.

Arcanemagus:
Looks like this bit would need to be expanded to also check `package.json`:
[...]

PR #1009, opened 2017-09-09
add specs for eslintIgnore key in package.json

skylize:
Adds specs for 1. eslintIgnore key in package.json is honored if no .eslintignore file found. 2. eslintIgnore key is ignored if .eslintignore file found. 3. linting occurs if no .eslintigore is found.


skylize:
This is definitely going to conflict with #1004. I'll be happy to resolve the conflicts after you merge that one, but wanted to go ahead and make this available. These tests show that #991 is in fact fixed.

spec/linter-eslint-spec.js
```diff
@@ -213,6 +215,46 @@ describe('The eslint provider for Linter', () => {
...
+      const tempPath = await copyFileToTempDir(path.join(eslintIgnoreDir, 'ignored.js'))
+      const tempDir = path.dirname(tempPath)
+
+      const editor = await atom.workspace.open(tempPath)
+      atom.config.set('linter-eslint.disableEslintIgnore', false)
+      await copyFileToDir(path.join(eslintIgnoreDir, '.eslintrc.yaml'), tempDir)
+
+      const messages = await lint(editor)
+      expect(messages.length).toBe(1)
+      expect(editor)
```
Arcanemagus: What's this line for?

spec/linter-eslint-spec.js
```diff
@@ -213,6 +215,46 @@ describe('The eslint provider for Linter', () => {
     })
   })
 
+  describe('when a file is not specified in .eslintignore file', async () => {
```
Arcanemagus: What's the point of this entire spec? Isn't it just testing that `copyFileToTempDir` works, and that linting works? (Both of which are tested/used elsewhere).

spec/linter-eslint-spec.js
```diff
@@ -213,6 +215,46 @@ describe('The eslint provider for Linter', () => {
...
+      const tempPath = await copyFileToTempDir(path.join(eslintIgnoreKeyDir, 'ignored.js'))
+      const tempDir = path.dirname(tempPath)
+
+      const editor = await atom.workspace.open(tempPath)
+      atom.config.set('linter-eslint.disableEslintIgnore', false)
+      await copyFileToDir(path.join(eslintIgnoreKeyDir, 'package.json'), tempDir)
+
+      const messages = await lint(editor)
+      expect(messages.length).toBe(0)
+      expect(editor)
```
Arcanemagus: What's this line for?

spec/fixtures/configs/eslintignorekey/.eslintignore
```diff
@@ -0,0 +1 @@
+**/filedoesntexist
```
Arcanemagus: This file doesn't appear to be used?

spec/linter-eslint-spec.js
skylize: This spec is to explicitly verify that linting occurs if no `.eslintignore` file is found before proceeding to test `eslintIgnore` key. There are already tests for what happens when `.eslintignore` is specified, but nothing to clearly verify what happens in it's absence. I guess you could think of this as a dummy-check for some of the other tests.

I personally did not believe the result from the other 2 specs I wrote until I wrote this one too. I thought I did something wrong until I could see clearly and for sure in a single test that "Yes, if you omit a `.eslintignore` you will get linting".

spec/linter-eslint-spec.js
skylize: Hahaha. I really have no idea. I can only assume it is remnant from some sort of rewrite/copy-paste something or other that I didn't notice since it doesn't break anything. Will 🔥

skylize:
@IanVS please see @Arcanemagus's question above about  "config root detection" . This seems outside the scope of this pull but worth discussing.

IanVS:
> but it almost seems that multiple .eslintignore files are allowed

What gives you that impression, @Arcanemagus?  

As for root detection, we might indeed need to revisit using the `.eslintignore` file location as a heuristic, due to a change I implemented in ESLint which now resolves the patterns in the file relative to the file location, rather than to CWD (which was the big reason we could use its location before as the assumed root).

> Do we need to expand our "config root detection" to handle a package.json file with an eslintIgnore key as a "forced root" like .eslintignore files?

I'm not sure.  From reading the code, it looks like eslint will only look for a `package.json` file within the `cwd`, so if we set `cwd` based on the presence of a `package.json` containing an `eslintIgnore` key, it feels like we're going about it backwards.  

I'm starting to wonder if we just need to do away with all the magic `cwd`-setting and always treat the top level that's open in atom as the project root.  Then at least we can clearly state that's how it works and people will know what to expect.

spec/linter-eslint-spec.js
Arcanemagus: Virtually every other spec (except the rare few that also deal with `.eslintignore` files) handles this.... _but_ they all do so under the state of `atom.config.set('linter-eslint.disableEslintIgnore', true)` which makes a good point for having a fully isolated spec like this, I'd be fine with leaving it in.

Arcanemagus:
> What gives you that impression, @Arcanemagus?

Apparently missing a sentence in what I remembered from reading the docs several days ago 😛. Checking them again there is a sentence "Only one `.eslintignore` file can be used at a time" which definitely says that behavior hasn't changed.

> I'm starting to wonder if we just need to do away with all the magic cwd-setting and always treat the top level that's open in atom as the project root.

On the surface I would love to just do that, there are edge cases to consider though:

* What about files open in Atom that aren't part of the current "project" in Atom?
* All the crazy path overrides and bizarre structures people have (like multi-project repos)

> This seems outside the scope of this pull but worth discussing.

@skylize Whoops, since this PR is just for adding specs, you're right. I was still thinking of this as the "add support for `eslintignore` keys" PR.

spec/fixtures/configs/eslintignorekey/.eslintignore
skylize: This file is required by spec `"will still lint the file if .eslintignore file"`.  This spec tries to lint `ignored.js` in this folder. When `.eslintignore` is found with a non-matching glob, `eslintIgnore` linting should occur, regardless of contents of `eslintIgnore` key.

skylize: I'm not really sure why this should be the case. But I'm just following the docs, which state 
> **If an .eslintignore file is not found** and an alternate file is not specified, eslint will look in package.json for an eslintIgnore key to check for files to ignore.

And sure enough, that seems to be the behavior we're getting.

Arcanemagus: Ah ha, so this is just checking that ESLint is behaving as it should.

spec/linter-eslint-spec.js
```diff
@@ -213,6 +215,44 @@ describe('The eslint provider for Linter', () => {
...
+      await copyFileToDir(path.join(eslintIgnoreDir, '.eslintrc.yaml'), tempDir)
+
+      const messages = await lint(editor)
+      expect(messages.length).toBe(1)
+      rimraf.sync(tempDir)
+    })
+  })
+
+  describe('when a file is specified in an eslintIgnore key in package.json', () => {
+    it('will still lint the file if .eslintignore file', async () => {
```
Arcanemagus: `will still lint the file if an .eslintignore file is present`

spec/fixtures/configs/eslintignorekey/.eslintignore
skylize: Yeah, I guess if you think about it it, this whole suite is redundant, because we're basically just saying `eslint` is doing for us what it says in the docs. Yet if you were to run this suite against the version of `linter-eslint` published to `apm`, it  would fail. So it seems like we should have it.

spec/linter-eslint-spec.js
skylize: 👍

skylize:
This should be ready to go. 

---
We have some flaky specs which I have seen many times failing due to unresolved promises.  When I first started looking at this code-base, I got failures almost every time when I first opened spec runner plus here-and-there at random, needing to hit reload to get a "valid test". Since I began watching for them so I can possibly fix those tests, it quit happening on my system.

When I asked about how to highlight code, it was in reference to some tests that look suspect for not guaranteeing to resolve a promise returned. I plan to open issues around those. However, on my last push (before this one) the Circle CI build failed due to 3 failing tests. I literally fixed the failed build by amending the commit message. If you look at the Circle CI build history, you should be able to pick out the offending tests. Let me know what they are, and I'll take a look.

Arcanemagus:
For reference, the failing specs you are talking about in [this build](https://circleci.com/gh/AtomLinter/linter-eslint/1049) are:
```
The eslint provider for Linter
  fixes errors
    it should fix linting errors
      timeout: timed out after 10000 msec waiting for spec promise to resolve
    it should not fix linting errors for rules that are disabled with rulesToDisableWhileFixing
      timeout: timed out after 10000 msec waiting for spec promise to resolve
  when an eslint cache file is present
    it does not delete the cache file when performing fixes
      timeout: timed out after 10000 msec waiting for spec promise to resolve
```

It looks like those 3 specs are all the ones that use the `makeFixes` function.

Arcanemagus:
For future reference btw, can you capitalize the first letter of your commits @skylize? I've been fixing it in the Squash and Merge 😉.

skylize:
@Arcanemagus 

You're welcome. Glad to help. Tests are actually a great way to jump in and help out on a project because their complexity is limited, they potentially explain what everything else is doing, and they tend to be neglected.

Those are not the specs I had in mind, but I'll look at those first. Better to fix the stuff proven to be flaky than the stuff that just looks suspicious.

Capitalize. Check.

PR #1014, opened 2017-09-16
Normalize all config paths

### LINKED ISSUE #985: Normalize paths in settings
This is not a bug, more an inconsistency with the settings.

I can set the location of my eslintrc file with a tilde like so:

```
~/.eslintrc
```

However, setting the Global Node Installation Path to

```
~/.config/yarn/global
```

does not work, but

```
/Users/mickyginger/.config/yarn/global
```
works fine

Arcanemagus:
Enhance all paths in the configurations so that they allow the use of `~` where appropriate.

Fixes #985.

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -69,7 +69,7 @@ export function findESLintDirectory(modulesDir, config, projectPath) {
     eslintDir = Path.join(cleanPath(config.advancedLocalNodeModules), 'eslint')
   } else {
     locationType = 'advanced specified'
-    eslintDir = Path.join(projectPath || '', config.advancedLocalNodeModules, 'eslint')
+    eslintDir = Path.join(projectPath || '', cleanPath(config.advancedLocalNodeModules), 'eslint')
   }
   if (isDirectory(eslintDir)) {
     return {
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -1,7 +1,7 @@
 'use babel'
 
 import Path from 'path'
-import fs from 'fs'
+import fs from 'fs-plus'
 import ChildProcess from 'child_process'
 import resolveEnv from 'resolve-env'
 import { findCached } from 'atom-linter'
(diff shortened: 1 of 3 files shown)


skylize:
Where is that config defined? Couldn't we just do something like this when we first grab it?
```
config.globalNodePath = globalNodePath.replace(/^~/, process.env.HOME)
```

Arcanemagus:
Right now the entire `linter-eslint` config is sent in with each job request. Re-working how the configuration is handled is on my to-do list, but for now I kept this PR minimal to work within the framework as it currently exists.

src/worker-helpers.js
```diff
@@ -13,6 +13,12 @@ const Cache = {
   LAST_MODULES_PATH: null
 }
 
+const cleanPath = (path) => {
```
IanVS: __*Preference*__

It might be good to drop a small docblock here to explain what this is doing to clean up the path.

src/worker-helpers.js
```diff
@@ -13,6 +13,12 @@ const Cache = {
...
+  const unTildeify = fs.normalize(path)
+  const unEnvVarify = resolveEnv(unTildeify)
```
IanVS: __*Preference*__

Having an ending of `ify` makes me think that these variables are functions, but it seems that they are actually just strings.  Honestly, I think you could just combine these three lines into:

```js
return resolveEnv(fs.normalize(path));
```

src/worker-helpers.js
```diff
@@ -45,7 +51,8 @@ export function findESLintDirectory(modulesDir, config, projectPath) {
   let locationType = null
   if (config.useGlobalEslint) {
     locationType = 'global'
-    const prefixPath = config.globalNodePath || getNodePrefixPath()
+    const configGlobal = config.globalNodePath && cleanPath(config.globalNodePath)
```
IanVS: __*Minor Suggestion*__

You could avoid the awkward `&&` here if `cleanPath` was changed to handle a falsey argument by just returning it unchanged or returning `null` or something.  It might not be a bad idea to add such a check to it no matter what, although I'm not sure what it does now with falsey values, maybe it already does.

src/worker-helpers.js
Arcanemagus: I think I was thinking of `ified`.

src/worker-helpers.js
Arcanemagus: I'm guessing it would have thrown a `TypeError` somewhere in there. If it's falsy I just made it return an empty string. (`null` would have caused errors when used in `path.*` functions.)

Issue #883, opened 2017-04-14
Better error messages for missing plugins


IanVS:
## Issue Type
Feature Request

## Issue Description
A common problem that `linter-eslint` users encounter is getting an error message like `Cannot find module 'eslint-plugin-xxxxx'`.  This is usually because they haven't installed it in the same location as the eslint that `linter-eslint` is using.  

We should catch these errors and give some intelligent instructions on how to fix the problem within the error message.  We already know what kind of ESLint they're using, so we should be able to say something like:

Local installation:
> The package `eslint-plugin-xxxxx` is not installed alongside ESLint.  Within your project directory, please run `npm install --save-dev eslint-plugin-xxxxx`.

Global installation:
> The package `eslint-plugin-xxxxx` is not installed alongside your global ESLint.  In a terminal, please run `npm install -g eslint-plugin-xxxxx`.

Fallback ESLint:
> You have configured your project to use `eslint-plugin-xxxxx`, but have not installed ESLint in your project.  The fallback version of ESLint installed with `linter-eslint` does not support eslint plugins.  Within your project directory, please run `npm install --save-dev eslint eslint-plugin-xxxxx`.

I think something like this would prevent a large portion of the issue reports we get on this project.

skylize:
I think this idea could certainly extend beyond plugins. This should just be "Better error messages".

The vast majority of runtime errors are due to improper setup on the user end. But in most of those cases we could automatically suggest a fix if we were actually parsing the Error messages.

Seems like the first step here is to start building a list of known errors, along with potential identifying features and causes of the problem. I'll start with your example:

> Failure:
> * **Plugin not found**
>
> Identifying features: 
> * Throws `Cannot find module` on `require`.
> * Module name matches on `/eslint-plugin-/`.
>
> Possible causes:
> * Plugin not installed.
> * ESLint running from different location than installed plugin.

PR #1015, opened 2017-09-17
Handle errors as a lint message

### LINKED ISSUE #387: linter-eslint is very loud when stuff is not installed
Currently, if linter-eslint tries to run, but stuff is missing from npm, it complains very loudly and repeatedly as I try to navigate through the project. It is really obstructive, and it quickly gets to the point where I either have to disable linter-eslint (which is a pain that I sometimes forget to undo) or I have to wait on `npm install`. (sometimes it just takes longer than I want)

One idea I have: if the exact same error message comes up repeatedly, it would only be sent as a notification one time. This could be accomplished by using `err.message` as a key and setting a timeout of 5 minutes or something before that message will be allowed to be shown again. (with the timeout length configurable)

To be clear, this is definitely a nit-pick on my part, especially since the solution is as simple as running `npm install`, but I think limiting how often these messages show up would be helpful.

Arcanemagus:
## Handle errors as a lint message

Translate all errors coming from ESLint into a lint message instead of throwing them out to be caught by Linter's generic error catching mechanism. This allows us to present them in a much cleaner manner and
gives the user immediate feedback on what is going wrong.

Also includes several cleanups related to error handling to get a more unified experience with errors.

### Before

Errors were only shown as an Atom error notification from Linter's generic catch-all:
image
Repeated lint requests would spam the message on the user's screen, providing a rather poor experience.

### After

Errors coming from ESLint will show as a message in Linter, with the expandable message details showing the full stack trace:
Linter message:
image
Full details:
image

Fixes #387.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -63,22 +63,6 @@ export async function sendJob(worker, config) {
   })
 }
 
-export function showError(givenMessage, givenDetail = null) {
-  let detail
-  let message
-  if (givenMessage instanceof Error) {
-    // mdn.io/Destructuring_assignment#Assignment_without_declaration
-    ({ stack: detail, message } = givenMessage)
-  } else {
-    detail = givenDetail
-    message = givenMessage
-  }
-  atom.notifications.addError(`[Linter-ESLint] ${message}`, {
-    detail,
-    dismissable: true
-  })
-}
-
 function validatePoint(textBuffer, line, col) {
   // Clip the given point to a valid one, and check if it equals the original
   if (!textBuffer.clipPosition([line, col]).isEqual([line, col])) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -149,10 +149,10 @@ export async function generateDebugString(worker) {
   return details.join('\n')
 }
 
-const generateInvalidTrace = async (
+const generateInvalidTrace = async ({
   msgLine, msgCol, msgEndLine, msgEndCol,
   eslintFullRange, filePath, textEditor, ruleId, message, worker
-) => {
+}) => {
   let errMsgRange = `${msgLine + 1}:${msgCol}`
   if (eslintFullRange) {
     errMsgRange += ` - ${msgEndLine + 1}:${msgEndCol + 1}`
@@ -276,10 +276,18 @@ export async function processESLintMessages(response, textEditor, showRule, work
         // This isn't an invalid point error from `generateRange`, re-throw it
         throw err
       }
-      ret = await generateInvalidTrace(
-        msgLine, msgCol, msgEndLine, msgEndCol,
-        eslintFullRange, filePath, textEditor, ruleId, message, worker
-      )
+      ret = await generateInvalidTrace({
+        msgLine,
+        msgCol,
+        msgEndLine,
+        msgEndCol,
+        eslintFullRange,
+        filePath,
+        textEditor,
+        ruleId,
+        message,
+        worker
+      })
     }
 
     return ret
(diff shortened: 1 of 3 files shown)


Arcanemagus:
@IanVS This _somewhat_ covers #883, but not fully which is why I didn't mark it as fixing that.

skylize:
Love the idea. Not a big fan of the full page stack trace. That looks even more invasive than throwing an Atom error. Is there a plan to be able to suppress the stack until user explicitly requests it?

Arcanemagus:
Updated the description to make it a bit clearer that the stack trace is only shown in the `description` of the message, an optionally displayed part that is meant to hold long form descriptions 😉.

src/helpers.js
```diff
@@ -150,10 +133,25 @@ export async function generateDebugString(worker) {
   return details.join('\n')
 }
 
-const generateInvalidTrace = async (
+export async function handleError(textEditor, error) {
```
Arcanemagus: Whoops, wrote this as `async` initially, but it doesn't actually need to be.

src/helpers.js
```diff
@@ -270,12 +270,6 @@ export async function processESLintMessages(response, textEditor, showRule, work
         ret.solutions = [linterFix]
       }
     } catch (err) {
-      if (!err.message.startsWith('Line number ') &&
-        !err.message.startsWith('Column start ')
-      ) {
-        // This isn't an invalid point error from `generateRange`, re-throw it
-        throw err
-      }
```
IanVS: __*Question*__

We are guaranteed that `ruleId` will be a string or `undefined`, right?  I only ask, because if it's truthy but something other than a string, `ruleURI()` will [throw a TypeError](https://github.com/jfmengels/eslint-rule-documentation/blob/cab02691eef796e7b18ec63467220330625ccd10/index.js#L20).

Arcanemagus: It comes straight from ESLint itself, so I'm not sure. I'll move that out of the `try` though just in case.

Arcanemagus:
The formatting issue would be a bug with Nuclide, it's not our fault they can't properly render messages 😛. (`description` is supposed to be optionally shown and can be quite large... it was tested with entire web pages rendered in there.)

Arcanemagus:
Looks like they made the same mistake in `atom-ide-ui`, filed https://github.com/facebook-atom/atom-ide-ui/issues/40 about it.

skylize:
I feel like Thrown Errors should be distinguished in some way from Linter Errors (even if message details not expanded). Perhaps just prepend "Atom" or "Thrown" to the Severity? So you would see "Atom Error" for thrown errors and just "Error" for linting errors.

Arcanemagus:
The errors aren't from Atom though? As stated in the message, the error is coming from ESLint.

Errors that are thrown outside of the handled sections are still bubbled up to Linter/Atom.

skylize:
Okay I understand. So it's an ESlint Error, but that name seems pretty meaningless since the linter errors are also from ESLint. "Thrown Error" or similar still makes to me here. I guess it's not hugely important. But if running in CLI, your output would be very very different from a thrown error vs a lint error. I feel like at least some little thing to clearly acknowledge this difference would be helpful.

Arcanemagus:
Do you have a better proposed text than the current `Error while running ESLint: ${message}`?

skylize:
Hmm. I typed a few ideas out, but none of them seem any better except this:
```
** Error while running ESLint ** : $(message)
```

Stars are a text version of highlighting. We're just trying to point out that "Hey, this is a special error." It's typically fatal and will prevent any further linting, so it should be easy to tell instantly that it's not a typo in this file file that's causing it.

Stars in the Description text seems sufficient to me as visually indicating the extra severity, actually better than my initial proposal.

PR #1016, opened 2017-09-17
Allow multiple rule directories

### LINKED ISSUE #321: Cannot pass in multiple rules dirs
It looks like eslintRulesDir only supports a single dir.

We have some rules supplied in an npm module, and some rules in our own repo, so need to pass in multiple rulesdirs.

For eslint on the command line it supports passing in --rulesdir multiple times

Arcanemagus:
Replace the `eslintRulesDir` option with a new `eslintRulesDirs` option that is an Array of directories that are passed to ESLint.

Fixes #321.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -62,6 +62,20 @@ module.exports = {
     this.subscriptions = new CompositeDisposable()
     this.worker = null
 
+    /**
+     * FIXME: Deprecated eslintRulesDir{String} option in favor of
+     * eslintRulesDirs{Array<String>}. Remove in the next major release,
+     * in v8.5.0, or after 2018-04.
+     */
+    const oldRulesdir = atom.config.get('linter-eslint.eslintRulesDir')
+    if (oldRulesdir) {
+      const rulesDirs = atom.config.get('linter-eslint.eslintRulesDirs')
+      if (rulesDirs.length === 0) {
+        atom.config.set('linter-eslint.eslintRulesDirs', [oldRulesdir])
+      }
+      atom.config.unset('linter-eslint.eslintRulesDir')
+    }
+
     const embeddedScope = 'source.js.embedded.html'
     this.subscriptions.add(atom.config.observe(
       'linter-eslint.lintHtmlFiles',
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -169,15 +169,13 @@ export function getCLIEngineOptions(type, config, rules, filePath, fileDir, give
     cliEngineConfig.ignorePath = ignoreFile
   }
 
-  if (config.eslintRulesDir) {
-    let rulesDir = cleanPath(config.eslintRulesDir)
+  cliEngineConfig.rulePaths = config.eslintRulesDirs.map((path) => {
+    const rulesDir = cleanPath(path)
     if (!Path.isAbsolute(rulesDir)) {
-      rulesDir = findCached(fileDir, rulesDir)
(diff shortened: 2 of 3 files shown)


src/worker-helpers.js
```diff
@@ -160,14 +160,17 @@ export function getCLIEngineOptions(type, config, rules, filePath, fileDir, give
...
+  if (config.eslintRulesDirs.length > 0) {
+    cliEngineConfig.rulePaths = []
+    config.eslintRulesDirs.forEach((path) => {
+      let rulesDir = resolveEnv(path)
+      if (!Path.isAbsolute(rulesDir)) {
+        rulesDir = findCached(fileDir, rulesDir)
+      }
+      if (rulesDir) {
+        cliEngineConfig.rulePaths.push(rulesDir)
+      }
```
IanVS: __*Preference*__

I think this could be a bit cleaner by using a more functional approach:

```js
function resolveRulesDir(path) {
    const rulesDir = cleanPath(path)
    if (!Path.isAbsolute(rulesDir)) {
        return findCached(fileDir, rulesDir)
    }
    return rulesDir
}

cliEngineConfig.rulePaths = config.eslintRulesDirs
    .map(resolveRulesDir)
    .filter(path => path)
```

src/main.js
```diff
@@ -62,6 +62,16 @@ module.exports = {
     this.subscriptions = new CompositeDisposable()
     this.worker = null
 
+    // FIXME: Remove this after a few editions
```
IanVS: __*Preference*__

It might not always be clear what "a few editions" means.  It might be better to say something like "Remove this in 2019" (hopefully that would be enough time. ;)

Also, a small comment here about what this block of code is doing might be good, just to give context on why it exists.  In fact, we change rules once in a while, and I wonder if it might make sense to create a separate file/function that translates old rules to new ones.  It might help with organization, but then again maybe it's not worth it.  Up to you.

Arcanemagus: > It might not always be clear what "a few editions" means.

I was thinking something along the lines of a few minor versions, but have nothing set in stone.

> I wonder if it might make sense to create a separate file/function that translates old rules to new ones.

I'd like to not be changing old options so often something like this is necessary, but the thought had crossed my mind.

IanVS: Yeah, in this case it's less about the number of versions, but the likelihood that users have opened Atom and run this code, which is why I thought a date constraint might make more sense.  NBD either way though.

skylize: How about this? 
```
/** FIXME: Deprecated eslintRulesDir{String} option in version 8.3.3 in favor of 
  * eslintRulesDirs{Array<String>}. Remove handling of eslintRulesDir after a few
  * minor updates or before 2019.
  */
```

Arcanemagus: I like that version @skylize, changed it a little bit in the upcoming change.

Arcanemagus:
@IanVS rebased and comments addressed.

src/worker-helpers.js
Arcanemagus: I sidestepped the whole issue of coming up with a name and just left it in the `map` 😉.

Arcanemagus:
Trying a close/re-open to wake up Travis CI... it _built_ the changes but hasn't told GitHub apparently.

Arcanemagus:
It looks like Travis CI just doesn't feel like talking to GitHub today, when AppVeyor (re)finishes I'll merge.

PR #1017, opened 2017-09-17
Check for remote files

### LINKED ISSUE #594: `Error: ENOENT: no such file or directory, uv_chdir` when using with nuclide and remote server
It gets thrown on every save

```
Error: ENOENT: no such file or directory, uv_chdir
    at Error (native)
    at Object.getRelativePath (/Users/111/.atom/packages/linter-eslint/lib/worker-helpers.js:113:11)
    at /Users/111/.atom/packages/linter-eslint/lib/worker.js:61:34
    at Emitter.emit (/Users/111/.atom/packages/linter-eslint/node_modules/process-communication/node_modules/sb-event-kit/lib/emitter.js:70:19)
    at /Users/111/.atom/packages/linter-eslint/node_modules/process-communication/node_modules/sb-communication/lib/main.js:22:23
    at /Users/111/.atom/packages/linter-eslint/node_modules/process-communication/node_modules/sb-communication/lib/main.js:21:24
    at Emitter.emit (/Users/111/.atom/packages/linter-eslint/node_modules/process-communication/node_modules/sb-event-kit/lib/emitter.js:68:19)
    at Communication.parseMessage (/Users/111/.atom/packages/linter-eslint/node_modules/process-communication/node_modules/sb-communication/lib/main.js:63:18)
    at process.ProcessCommunication.messageCallback (/Users/111/.atom/packages/linter-eslint/node_modules/process-communication/lib/communication.js:35:27)
    at emitTwo (events.js:87:13)
```

Arcanemagus:
Add several sanity checks before moving on to attempting to lint the file:

* Check that the `TextEditor` is valid
* Check that the path for the file is not empty
* Check that the path isn't a remote path

If a remote path is found, show a message to the user that linting isn't possible.

Blocked on #1015 as it uses the Error handling functionality introduced in that branch. Just putting this up here so it can get reviewed.

Fixes #594.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -188,11 +188,26 @@ module.exports = {
       scope: 'file',
       lintsOnChange: true,
       lint: async (textEditor) => {
-        const text = textEditor.getText()
-        if (text.length === 0) {
-          return []
+        if (!atom.workspace.isTextEditor(textEditor)) {
+          // If we somehow get fed an invalid TextEditor just immediately return
+          return null
         }
+
         const filePath = textEditor.getPath()
+        if (!filePath) {
+          // The editor currently has no path, we can't report messages back to
+          // Linter so just return null
+          return null
+        }
+
+        if (filePath.includes('://')) {
+          // If the path is a URL (Nuclide remote file) return a message
+          // telling the user we are unable to work on remote files.
+          const message = 'Remote file open, impossible to run ESLint.'
+          helpers.handleError(textEditor, new Error(message))
+        }
+
+        const text = textEditor.getText()
 
         if (!helpers) {
           helpers = require('./helpers')
--- a/src/main.js
+++ b/src/main.js
@@ -203,8 +203,10 @@ module.exports = {
         if (filePath.includes('://')) {
           // If the path is a URL (Nuclide remote file) return a message
           // telling the user we are unable to work on remote files.
-          const message = 'Remote file open, impossible to run ESLint.'
-          helpers.handleError(textEditor, new Error(message))
+          return helpers.generateUserMessage(textEditor, {
+            severity: 'warning',
+            excerpt: 'Remote file open, linter-eslint is disabled for this file.',
+          })
         }
 
         const text = textEditor.getText()
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -143,6 +143,9 @@ export async function generateDebugString(worker) {
  * Turn the given options into a Linter message array
  * @param  {TextEditor} textEditor The TextEditor to use to build the message
  * @param  {Object} options    The parameters used to fill in the message
+ * @param  {string} [options.severity='error'] Can be one of: 'error', 'warning', 'info'
+ * @param  {string} [options.excerpt=''] Short text to use in the message
+ * @param  {string|Function} [options.description] Used to provide additional information
  * @return {Array}            Message to user generated from the parameters
  */
 export function generateUserMessage(textEditor, options) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -139,19 +139,42 @@ export async function generateDebugString(worker) {
   return details.join('\n')
 }
 
-export function handleError(textEditor, error) {
-  const { stack, message } = error
-  // Only show the first line of the message as the excerpt
-  const excerpt = `Error while running ESLint: ${message.split('\n')[0]}.`
+/**
+ * Turn the given options into a Linter message array
+ * @param  {TextEditor} textEditor The TextEditor to use to build the message
+ * @param  {Object} options    The parameters used to fill in the message
+ * @return {Array}            Message to user generated from the parameters
+ */
+export function generateUserMessage(textEditor, options) {
+  const {
+    severity = 'error',
+    excerpt = '',
+    description,
+  } = options
   return [{
-    severity: 'error',
+    severity,
     excerpt,
-    description: `<div style="white-space: pre-wrap">${message}\n<hr />${stack}</div>`,
+    description,
     location: {
       file: textEditor.getPath(),
       position: generateRange(textEditor),
     },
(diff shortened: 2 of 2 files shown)


skylize:
Should possibility of remote file checking be on our radar?

Arcanemagus:
For simple linters that _just_ depend on the file contents it's certainly a possibility, but `eslint` depends on a large amount of state in the rest of the project surrounding the file being linted. This means we need full access to the file system surrounding the file when linting it which isn't possible with a remote file.

skylize:
> For simple linters that just depend on the file contents it's certainly a possibility, but eslint depends on a large amount of state in the rest of the project surrounding the file being linted.

I think treating remote files as a single file with no project context would be a reasonable trade-off. If we get an .eslintrc global fallback working, it seems quite reasonable to use that fallback on new files that have never been saved. We could then simply treat remote files as new/never-saved.. Worth thinking about at least I think.

Arcanemagus:
> I think treating remote files as a single file with no project context would be a reasonable trade-off.

The most that could be done is ESLint's syntax check, and even then I'm not sure if reporting fixes back would even work as I don't know if the messages can be associated back to the file in any meaningful way.

> We could then simply treat remote files as new/never-saved.

Linter shouldn't even be triggering us on unsaved files in the first place 😉.

skylize:
> The most that could be done is ESLint's syntax check, and even then I'm not sure if reporting fixes back would even work as I don't know if the messages can be associated back to the file in any meaningful way.

A basic syntax check is still a very useful feature of having a linter handy. If I'm editing a remote or unsaved file and I have choice between linter only does basic syntax check or linter ignores the file, I would definitely take the syntax check. As for whether we can report the findings, that's I don't know about, but if there is a reasonable way it would be a good feature.

> >We could then simply treat remote files as new/never-saved.

> Linter shouldn't even be triggering us on unsaved files in the first place 😉.

Yes. This statement was based on the premise of first having useful treatment for unsaved files, which we currently do not. 

I just think it's worth having this potential use-case in the back of our minds in case someone stumbles onto a good way to do it or wakes up from a dream about linting with the perfect solution. Hahaha

Arcanemagus:
> Editing a remote or unsaved file

There's a _slim_ possibility a remote file would work, but as there is no path at all for an unsaved file there is no way to report messages to it (This is why `linter` doesn't support unsaved files currently, the API would have to be modified to pass a reference to an associated `TextEditor` on _every message_).

> Someone stumbles onto a good way to do it or wakes up from a dream about linting with the perfect solution

It's been over a year with no better solutions presented in that issue, so I went with the safest one 😉.

src/main.js
```diff
@@ -188,11 +188,26 @@ module.exports = {
...
+        if (!filePath) {
+          // The editor currently has no path, we can't report messages back to
+          // Linter so just return null
+          return null
+        }
+
+        if (filePath.includes('://')) {
+          // If the path is a URL (Nuclide remote file) return a message
+          // telling the user we are unable to work on remote files.
+          const message = 'Remote file open, impossible to run ESLint.'
```
IanVS: Perhaps a slightly-more-accurate message might be something like:

"Remote file open, linter-eslint is therefore disabled for this file."

Since it sounds like the Facebook folks have may their own eslint integration for remote files.

src/main.js
```diff
@@ -188,11 +188,26 @@ module.exports = {
       scope: 'file',
       lintsOnChange: true,
       lint: async (textEditor) => {
-        const text = textEditor.getText()
-        if (text.length === 0) {
-          return []
```
IanVS: __*Question*__

Did you mean to remove this check?

Arcanemagus: Hmmm, nope. I _meant_ to move it down below the path checks along with the `getText` call. 🐛

src/main.js
Arcanemagus: Huh, it would have to be some sort of primitive LSP to work at all. They did start Nuclide before Microsoft published the LSP protocol.

src/main.js
Arcanemagus: Oh, actually, yes I did mean to remove it. With it in there we are saying that an empty file is completely valid, but we have _no_ idea what a user's configuration is, or what plugins they have or rules they have enabled. We have no way of knowing whether an empty file is considered "valid" under their configuration so we should remove this optimization and do the full ESLint run even on an empty file.

Arcanemagus:
@IanVS fixed it up a bit 😉.

tim-soft:
Just for clarity, there's definitely no way we can lint projects _inside atom_ via Nuclide + Watchman?

Arcanemagus:
All that we have for files opened like that is the raw text, since ESLint depends on external configuration files which we have no way to access with these remote files there really isn't anything that _could_ be done other than a basic syntax check possibly.

Just because you can see the text in Atom doesn't mean the hacked together way that Nuclide has done that works for _anything_ other than Nuclide.

tim-soft:
Even if there were some ssh tunnel to the remote project dir?

Arcanemagus:
You could use `sshfs` or something similar and open the directory "locally" so it is accessible as part of the filesystem, the remote URL files aren't usable though.

tim-soft:
I guess running some kind of eslint server to report back linting results to atom from the remote project would be inadvisable as well?

Arcanemagus:
An ESLint [Language Server](https://github.com/Microsoft/language-server-protocol) is certainly _possible_, but none exists currently that I know of. From what I understand there is nothing that _prevents_ a language server from running on a remote host, I don't think any of the current adapters into Atom work like that though.

tim-soft:
Ah, I see. Making an eslint-js language client seems like the only way something like this would integrate with atom super nicely. https://github.com/atom/atom-languageclient

Thanks for the clarification

PR #1018, opened 2017-09-17
Add option to silence fixable rules while typing

### LINKED ISSUE #894: option to silence all fixable rules while typing
## Issue Type

Feature Request

## Issue Description

The silence specific rules while typing option is amazing but difficult to keep up to date with .eslintrc changes and, especially, to synchronise between team members.
If eslint exposes a list of "fixable" rules (or otherwise), can a checkbox that enable all the fixable rules to be silenced and just fixed on save?

Arcanemagus:
Add an option that silences fixable rules while typing. The list of fixable rules is populated from ESLint itself, and only sent back to the main package code when it updates to reduce message size. This means that this will only start silencing rules after the first lint rule. Since a lint is triggered on open by default this _should_ be largely transparent to the user.

Fixes #894.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,12 @@
       "items": {
         "type": "string"
       }
+    },
+    "ignoreFixableRulesWhileTyping": {
+      "title": "Ignore fixable rules while typing",
+      "description": "Have the linter ignore all fixable rules during linting when editing a document. The list is automatically updated on each lint job, and requires at least one run to be populated.",
+      "type": "boolean",
+      "default": false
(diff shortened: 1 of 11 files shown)


Arcanemagus:
Just found a major bug in this, and I'm expanding the tests to check for it. (This was silencing _all_ rules due to some unfinished code, but the specs weren't showing this!)

Arcanemagus:
Thanks for fixing that up and the detailed testing @IanVS! ❤️

PR #1019, opened 2017-09-17
Fix deprecation warning in Atom v1.21.0

Arcanemagus:
Atom has started showing a deprecation warning for the Atom v1.18.0 style check done for the context menu. Remove this check making the context menu now only show on Atom v1.19.0 and above.

image


Arcanemagus:
Pushing this through without review as this is causing deprecation notices to users and is a tiny change.

PR #1040, opened 2017-10-25
Add checkboxes to issue template.

richardpringle:


Arcanemagus:
I'm not opposed to this, I just worry that issues will then start having `0/3` all through the list and it won't actually make a difference as to whether the requested information was actually filled in or not.

Thoughts @IanVS?

richardpringle:
haha, just saw "checklist" in the header and thought I would throw it in there. 

As someone who's moderated hoards of GitHub issues, you're absolutely right that most people won't fill them in and this very little value added here.

skylize:
@Arcanemagus  +1 to try it and see.

PR #1042, opened 2017-10-30
Organize config options

IanVS:
This is an attempt to introduce some organization into our config settings.  Currently, they are ordered alphabetically by config name, which is to say, they're in completely random order.  That's pretty confusing for new users, trying to understand our 15 different options.

In an attempt to organize them and add some clarity, I've broken out three new sections, and grouped the remaining at the top (with boolean options above text options, because I think that looks nicer).  

The groups are:

- Global (global ESLint settings)
- Disabling (ways in which ESLint warnings can be turned off)
- Advanced (less common options that nearly all users can leave at the defaults)

The method of grouping settings is a little unfortunate, because it extends down into our code and specs.  I believe that I have made the necessary changes, but they should be reviewed.  I recommend searching the codebase for the names of each setting that is being grouped, and verifying that they are being correctly referenced.

If this PR is accepted, I will open another PR updating our README with some more in-depth explanations of the groupings and settings themselves.

image


IanVS:
I'd love for @skylize to take a look at this as well.

skylize:
I like this on first glance. I'll take a deeper look and get back to you.

README.md
```diff
@@ -53,8 +53,7 @@ eslint --init
 ```
 
 Alternatively you can create the `.eslintrc` file by yourself. It is a good
-idea to have a look at the [Get Started With ESLint](http://devnull.guru/get-started-with-eslint/)
-blog post by [IanVS](https://github.com/IanVS) and [the ESLint documentation](http://eslint.org/docs/user-guide/configuring),
+idea to have a look at the [the ESLint documentation](http://eslint.org/docs/user-guide/configuring),
```
Arcanemagus: `the the ESLint` 😛

package.json
```diff
@@ -87,31 +35,117 @@
...
+      "properties": {
+        "disableEslintIgnore": {
+          "title": "Don't use .eslintignore files",
+          "type": "boolean",
+          "default": false,
+          "order": 1
+        },
+        "disableFSCache": {
+          "title": "Disable FileSystem Cache",
+          "description": "Paths of node_modules, .eslintignore and others are cached",
```
Arcanemagus: I know you didn't change the wording here, but since we are touching it we might as well update a few things. Can you change this to "... others are normally cached"? The way it's currently worded the _description_ seems to indicate enabling this option enables caching, when the title (and behavior) do the exact opposite.

package.json
```diff
@@ -87,31 +35,117 @@
...
       "items": {
         "type": "string"
-      }
+      },
+      "order": 3
     },
-    "rulesToSilenceWhileTyping": {
-      "title": "Silence specific rules while typing",
-      "description": "Useful when Atom fixes errors on save like `no-trailing-spaces` or `eol-last`.",
+    "eslintRulesDirs": {
```
Arcanemagus: Is there any reason this isn't in the advanced section?

package.json
```diff
@@ -87,31 +35,117 @@
...
+          "order": 1
+        },
+        "disableFSCache": {
+          "title": "Disable FileSystem Cache",
+          "description": "Paths of node_modules, .eslintignore and others are cached",
+          "type": "boolean",
+          "default": false,
+          "order": 2
+        },
+        "showRuleIdInMessage": {
```
Arcanemagus: One could make the argument this should go in the "general uncategorized" area, but I think it's fine leaving this down here.

package.json
```diff
@@ -87,31 +35,117 @@
...
+          "order": 4
+        }
       }
     },
-    "ignoreFixableRulesWhileTyping": {
-      "title": "Ignore fixable rules while typing",
-      "description": "Have the linter ignore all fixable rules during linting when editing a document. The list is automatically updated on each lint job, and requires at least one run to be populated. Only supported when using ESLint v4+.",
-      "type": "boolean",
-      "default": false
+    "advanced": {
```
Arcanemagus: We can add a `collapsed` attribute to make this section collapsed by default.

Although it doesn't seem to be _documented_, it's [tested in the specs](https://github.com/atom/settings-view/blob/7689e9e04ae5b1f040a071611d372570ed518c50/spec/settings-panel-spec.coffee#L205).

src/migrate-config-options.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+]
+
+function migrateConfigOptions() {
+  /**
+   * FIXME: Deprecated eslintRulesDir{String} option in favor of
+   * eslintRulesDirs{Array<String>}. Remove in the next major release,
+   * in v8.5.0, or after 2018-04.
+   */
+  const oldRulesdir = atom.config.get('linter-eslint.eslintRulesDir')
```
Arcanemagus: Do we want to integrate this into the data structure above as a `migrationFunc` or something like that? If it exists it gets called instead of the default migration function.

src/migrate-config-options.js
```diff
@@ -0,0 +1,87 @@
...
+    const rulesDirs = atom.config.get('linter-eslint.eslintRulesDirs')
+    if (rulesDirs.length === 0) {
+      atom.config.set('linter-eslint.eslintRulesDirs', [oldRulesdir])
+    }
+    atom.config.unset('linter-eslint.eslintRulesDir')
+  }
+
+  // Copy old settings over to the new ones, then unset the old setting keys
+  directMoveMigrations.forEach((migration) => {
+    const oldSetting = atom.config.get(migration.old)
```
Arcanemagus: I think it might be worth it to do a single `atom.config.get('linter-eslint')` and then do our further processing over the return from that instead of calling `atom.config.get(key)` on each value, _especially_ since we are migrating essentially every setting we have.

package.json
Arcanemagus: Filed https://github.com/atom/atom/issues/16030 about getting it officially documented 😉.

skylize:
* `eslintrcPath` --> `Global`. Typical use would be with a global eslint. 

* Maybe `fixOnSave` --> `Advanced`?

    * I think it's not that easy for people to get this behaving right. It certainly should _not_ be the very first setting, encouraging everybody to dive in asap.

    * Or maybe a `Fix On Save` section with:
        * `fixOnSave`
        * `ignoreFixableRulesWhileTyping`
        * `rulesToDisableWhileFixing`
        * `rulesToSilenceWhileTyping`

* `disableEslintIgnore` --> `Disabling`. Seems fairly innocuous and self-explanatory.

* `showRuleIdInMessage` --> `general`. Nothing even slightly advanced here.

Arcanemagus:
> `eslintrcPath` --> `Global`. Typical use would be with a global eslint.

👍

> Or maybe a `Fix On Save` section

I rather like the idea of grouping everything related to fixing on save together as well.

> `disableEslintIgnore` --> `Disabling`. Seems fairly innocuous and self-explanatory.

That section is about turning _off_ messages, not turning them on 😉. In any case I'd say this is a rather advanced option that should only rarely be enabled (it was added for ESLint itself development...).

IanVS:
Thanks for the feedback guys!  I'll take a crack at some cleanup tonight.  

>>disableEslintIgnore --> Disabling. Seems fairly innocuous and self-explanatory.
>
>That section is about turning off messages, not turning them on 😉. In any case I'd say this is a rather advanced option that should only rarely be enabled (it was added for ESLint itself development...).

Yeah, the section name probably isn't great, but it's more about disabling ESLint warnings for various situations.  Perhaps it would be better named `Disabling Linting` or `Disabling Warnings` or something like that.

package.json
IanVS: I don't really understand why someone would want to turn that off.  Do you know off-hand the reason for the option to exist at all?

package.json
IanVS: I suppose it could be.  One of the big things that made ESLint different from JSHint was that you could write your own rules and they could be included at execution time.  Once-upon-a-time the only way was by giving a directory of rules files.  Now of course there are plugins, but I don't have a great sense for how many users out there still use custom rules in a rules directory.

package.json
IanVS: I was planning to take another pass at the descriptions after this PR, I think there are a few things that could probably be clarified.  I'll make this change now though since you've pointed it out and we may as well.

src/migrate-config-options.js
IanVS: Yeah I can play around with that.

package.json
skylize: It would make the output cleaner and easier to read at the cost of being much harder to look up rules in the docs or write inline disable rules. _I_ would never turn this off, but I can see the motivation. I think it's reasonable to argue for removing it entirely.

IanVS: Looks like it was added in https://github.com/AtomLinter/linter-eslint/commit/72cc3cb24be8e488030bb9d8eccae1cec0f7355c, I was hoping for a bit more context as to _why_, but no luck.

Arcanemagus: If I remember correctly he wanted to remove it completely when doing the rewrite, but I made him keep it around (this was before the rule ID even linked to the relevant page, so it was very helpful for googleing an error). The compromise was an option that defaulted enabled.

package.json
Arcanemagus: Interesting, I guess I just assumed that anyone writing their custom rules would format them as a plugin so they would work like all other custom rules.

package.json
skylize: When I said you could argue to remove it, I meant you could remove the setting. Removing the rule id for everybody would be incredibly mean.

Arcanemagus: Oh personally I would be for removing the setting, but there are people out there that don't like the "noise".

IanVS:
> `showRuleIdInMessage` --> `general`. Nothing even slightly advanced here.

I guess.  I just wanted to hide it because I think it's a dumb option.  :-p

package.json
```diff
@@ -87,31 +21,140 @@
...
-      "items": {
-        "type": "string"
+    "global": {
+      "type": "object",
+      "order": 4,
+      "title": "Global ESLint",
+      "properties": {
+        "useGlobalEslint": {
+          "title": "Use global ESLint installation",
+          "description": "Make sure you have it in your $PATH",
```
skylize: Could we add something here to emphasize that local installation is the suggested usage?

IanVS: As I mentioned to @Arcanemagus in one of the previous comments, I'd like to save wording tweaks for another PR.  There is a lot that I would like to do to help explain and clarify our settings, but this PR is really just about reorganizing them.

skylize: Fair enough. 👍

src/migrate-config-options.js
```diff
@@ -0,0 +1,104 @@
...
+ *   the `new` key.
+ *
+ * Functions:
+ *   These have a `type` of `function`, and a `migrationFunc`, which takes the
+ *   current linter-eslint atom config as an argument, and can act on it however
+ *   it needs to.
+ */
+const migrations = [
+  {
+    type: 'direct-move',
```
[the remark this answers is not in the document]
skylize: I think the direct move is more simple and clear for this, but it is a bit annoying to need a function type also because of the not-direct migration from earlier. Since we're running a function anyway, I guess I agree it seems better to go ahead and handle it inside the function.

Definitely agree with the rename.

src/migrate-config-options.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+function migrateConfigOptions() {
+  const linterEslintConfig = atom.config.get('linter-eslint')
+  migrations.forEach((migration) => {
+    if (migration.moves && Array.isArray(migration.moves)) {
+      // Copy old settings over to the new ones, then unset the old setting keys
+      migration.moves.forEach((move) => {
+        const oldSetting = linterEslintConfig[move.old]
+        if (oldSetting !== undefined) {
+          atom.config.set(move.new, oldSetting)
```
Arcanemagus: Missing the `linter-eslint` package name on the settings key.

src/migrate-config-options.js
```diff
@@ -0,0 +1,104 @@
...
+          atom.config.unset(move.old)
```
Arcanemagus: Missing the `linter-eslint` package name on the settings key.

src/migrate-config-options.js
```diff
@@ -0,0 +1,104 @@
...
+]
+
+/*
+ * This function can be called when linter-eslint first activates in order to
+ * ensure that the user's settings are up-to-date with the current version of
+ * linter-eslint.  Ideally, we would call this only when upgrading to a new
+ * version.
+ */
+function migrateConfigOptions() {
+  const linterEslintConfig = atom.config.get('linter-eslint')
```
Arcanemagus: 🤔 What about only grabbing this in the `forEach` if it's not already set?

It's not quite as clean, but it _does_ save grabbing the entire settings if there are no migrations to perform.

package.json
```diff
@@ -87,31 +21,140 @@
...
+          "title": "ESLint Rules Directories",
+          "description": "Specify a comma separated list of directories for ESLint to load rules from.",
+          "type": "array",
+          "default": [],
+          "items": {
+            "type": "string"
+          },
+          "order": 4
+        },
+        "advancedLocalNodeModules": {
```
Arcanemagus: 🤔 Since we are already migrating this to a new key, what about renaming it to just `localNodeModules`?

src/migrate-config-options.js
skylize: Based on the way this is laid out, I don't see any situation where your suggestion would somehow prevent grabbing the config. The `if` check is just so he can handle `eslintRulesDirs` differently than the straight renames.

Arcanemagus: I'm thinking about later, when we completely deprecate and remove the current migrations. Unless we are going to keep them around forever.

skylize: Well as long as we're messing around with it, I think this could be much more clear by increasing the symmetry between different types of migrations, as in they should _all_ be function calls, which also removes the need to check if `moves` exists. So combining that with your "maybe don't grab config", that would probably look something like this.

```
const migrations = [
  {
    added: 'January, 2018',
    description: 'Organized config settings into sections',
    migrate: (config) => {
      const moves = [ /* ... list of moves ... */ ]
      moves.forEach((move) => {
        const oldSetting = config[move.old]
        if (oldSetting !== undefined) {
          atom.config.set(move.new, oldSetting)
          atom.config.unset(oldSetting)
        }
      })
    }
  },
  {
    added: 'September, 2017'
    // ... 
  }
]

// If nothing to migrate don't grab config
const linterEslintConfig = migrations.length 
  ? atom.config.get('linter-eslint')
  : null

// If nothing to migrate, no problem that config is
// null because this is a no-op.
migrations.forEach(migration => migration.migrate(config))
```

Arcanemagus: I'd move that inner `(move) =>` function out to it's own definition since it will likely be used elsewhere, but that does look cleaner.

spec/migrate-config-options-spec.js
```diff
@@ -0,0 +1,37 @@
+'use babel'
+
+import migrateConfigOptions from '../src/migrate-config-options'
+import makeSpy from './make-spy'
+
+describe('migreateConfigOptions()', () => {
```
skylize: typo her

IanVS:
@Arcanemagus I fixed the bug in the migration code and added some specs.  I opted not to refactor as suggested in https://github.com/AtomLinter/linter-eslint/pull/1042#discussion_r164627764, mostly because it ended up making things more complicated.  Specifically, I foresee that most migrations we do will be simple moves, and having a `moves` array is pretty dead simple.  Also, testing became a bit more troublesome when using a `migrate` function even for simple moves, because then I would need to also export the helper function and honestly it just didn't all seem worth it.  If you think it's a blocker, then I can re-address, but I'm kinda at the point where I'd just like to be done with this.  :)

spec/migrate-config-options-spec.js
IanVS: damn…

IanVS: I do think it's funny that you also had a typo, in pointing out my typo:

> typo her

skylize:
It seems to get the job done. I'm not gonna stand in the way. We can always iterate. (If we manage to get fun-functional merged in, it all gets changed anyway to fit better with the newer config subscription setup.)

PR #1067, opened 2018-01-07
Use getRules and rule URL metadata where available

Arcanemagus:
Refactor the worker to send the full metadata instead of just the fixable rules, allowing additional functionality to be implemented in the main package code.

Rule gathering from the ESLint instance has been expanded to take advantage of the new [`CLIEngine#getRules`](https://github.com/eslint/eslint/pull/9782) method if it is available.

The rule URL is now pulled directly from the metadata ([`rule.meta.docs.url`](https://github.com/eslint/eslint/pull/9788)) if it is available, falling back to the current method of using `eslint-rule-documentation` if the rule doesn't specify this for itself.

These new features were added in ESLint v4.15.0.

----

This also refactors how rules are handled a bit so the following behavior is how it works now:

* On each lint in the **worker**:
  * The list of known rules is checked against what ESLint just used to lint
  * If it changed, the entire list is sent back to the main thread (later optimization to only send diff? We _have_ had issues with total message size...)
  * If it didn't change, only the regular results are sent back
* In the **main thread** (`helpers.js`):
  * When a response from a job contains updated rules, an instance of the new `Rules` object is updated with the new set of rules
  * Whenever a lint job is requested and the user has "Ignore fixable rules" enabled this `Rules` instance is queried for the current list of fixable rules.
  * Whenever messages are being processed the `Rules` instance is queried for a URL for the rule's documentation.


Arcanemagus:
If you would prefer that the refactor work is split out to a separate PR from the `getRules` / URL changes just let me know @IanVS.

IanVS:
@Arcanemagus I hate to say it, but yeah if you could split out the refractor work that would be helpful.  If you want to keep it in this PR, that's fine, but having a separate commit would be good I think.

src/helpers.js
```diff
@@ -319,6 +320,39 @@ export async function processESLintMessages(messages, textEditor, showRule, work
...
+      fixableRules.add(rule)
+    }
+  })
+}
+
+/**
+ * Process the updated rules into the local Map and call further update functions
+ * @param  {Array} updatedRules Array of Arrays of the rule and properties
+ */
+export function updateRules(updatedRules) {
```
IanVS: __*Preference*__

I'm not sure `update` is the best name here, because it kind of makes it sound like like the old rules are being kept and just somehow modified.  I know this is niggling, but I think I'd prefer a name like `replaceRules` in this case.

src/helpers.js
```diff
@@ -319,6 +320,39 @@ export async function processESLintMessages(messages, textEditor, showRule, work
   }))
 }
 
+/**
+ * Update the list of fixable rules
+ */
+function updateFixableRules() {
```
IanVS: What do you think about this function taking `fixableRules` as an argument, to minimize the side effects and avoid mutating something outside of its own scope?  I realize that if you pass it in and modify it, that's still technically a side effect, but it feels slightly cleaner.

src/helpers.js
```diff
@@ -8,6 +8,7 @@ import cryptoRandomString from 'crypto-random-string'
 // eslint-disable-next-line import/no-extraneous-dependencies, import/extensions
 import { Range } from 'atom'
 
+const lintRules = new Map()
```
IanVS: It feels to me like if we go this route, `lintRules` should be a property on a `class`, and `updateFixableRules`, `updateRules`, and `getRules` should be methods on that class.  Normally I'd advocate for a more functional style, but I find maps to be difficult to work with in pure functions.

src/worker-helpers.js
```diff
@@ -184,3 +184,56 @@ export function getCLIEngineOptions(type, config, rules, filePath, fileDir, give
...
+ * @param  {[type]} currentRules A Map of the current rules
+ * @return {boolean}             Whether or not there were changes
+ */
+export function didRulesChange(currentRules, newRules) {
+  let rulesChanged = false
+  const currentRuleIds = new Set(currentRules.keys())
+  const newRuleIds = new Set(newRules.keys())
+
+  // Check for new rules added since the last time we sent currentRules
+  const newRulesIds = new Set(newRules.keys())
```
IanVS: Is it intentional to have `newRuleIds` and `newRulesIds`?  If so, that's pretty confusing, IMHO.

I like the approach to comparing Sets laid out in this SO answer: https://stackoverflow.com/a/44827922/1435658

```js
areSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
```

src/worker.js
```diff
@@ -9,64 +9,19 @@ import isConfigAtHomeRoot from './is-config-at-home-root'
...
-}
-
 function lintJob({ cliEngineOptions, contents, eslint, filePath }) {
   const cliEngine = new eslint.CLIEngine(cliEngineOptions)
   const report = cliEngine.executeOnText(contents, filePath)
-  // Use the internal (undocumented) `linter` instance attached to the cliEngine
-  // to check the loaded rules (including plugin rules) and update our list of fixable rules.
-  updateFixableRules(cliEngine.linter)
+  const rules = Helpers.getRules(cliEngine)
+  sendRules = Helpers.didRulesChange(rulesMetadata, rules)
```
IanVS: __*Preference*__

I know this wasn't created in this PR, but as a boolean constant, let's name this something like `shouldSendRules`.

src/worker.js
```diff
@@ -111,7 +66,10 @@ module.exports = async () => {
         messages: report.results.length ? report.results[0].messages : []
       }
       if (sendRules) {
-        response.fixableRules = Array.from(fixableRules.keys())
+        // You can't emit Maps, convert to Array of Arrays to send back.
+        response.updatedRules = []
+        rulesMetadata.forEach((props, rule) =>
+          response.updatedRules.push([rule, props]))
```
IanVS: Doesn't `Array.from(rulesMetadata)` do what you want here?  And if not, can we use [`rulesMetadata.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries)?

src/helpers.js
Arcanemagus: I went with `update` since as far as anything outside this function is concerned it _is_ an update. The fact that it happens to clear the current contents and then fill in new ones is an implementation detail.

IanVS: I see your reasoning.  Personally, if I call `updateThings(newThing)`, it isn't immediately clear as a consumer of the function whether all the old things will be gone or not.  I might think that I'm just adding a new thing to the existing things.  I guess all I'm saying is that `update` is vague, and I find `replace` to be clearer, but I'm only at a 3 of 10 on the [SSOGAF](http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck).

skylize: It's certainly not ideal that it clears the old rules in place as a side-effect before updating them. But the general concept of full replacement as an update is totally valid, and the foundation of immutability and great tools like Redux.

src/worker-helpers.js
Arcanemagus: It's definitely confusing at best, and that's much cleaner. I'll switch it to something like that 😉.

src/worker.js
Arcanemagus: Can do.

src/helpers.js
IanVS: I have no issue with doing it that way, just the _naming_.

src/worker.js
Arcanemagus: Honestly I didn't even think to try it. `Array.from(<some Map>)` does _exactly_ this. 👍

src/helpers.js
Arcanemagus: Updating in place gets complex since then you have to determine what you need to remove, and what you need to insert. Since there is nothing actually _tied_ to the contents here a simple replacement works just fine.

I'd say I'm at around a 4 or 5 on the SSOGAF 😛.

IanVS: Haha, you called it a _replace_ment.  😝

src/rules.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+export default class Rules {
+  /**
+   * Instantiates a Rules object, optionally with an existing list of rules
+   * @param {Array} newRules Array of Arrays of the rule and properties
+   */
+  constructor(newRules) {
+    this.rules = new Map()
+    if (newRules) {
+      this.updateRules(newRules)
```
IanVS: I'm no `Map` expert, but couldn't this be done in one step?  Something like:

```js
this.rules = new Map(newRules)
```

Arcanemagus: Since the constructor will never be called with a parameter except in specs, sure. If this was a "real" class though we would want validation in there that whatever `newRules` is is something that actually makes sense to instantiate with.

IanVS:
I'd still prefer `replaceRules` to `updateRules`, but I won't block a merge over it.  :)

PR #1069, opened 2018-01-07
Remove definition of warnIgnored in CLIEngine options

Arcanemagus:
This flag has no meaning to the `CLIEngine` constructor, it is only valid as an optional third argument to `executeOnText`. When this was mistakenly added in https://github.com/AtomLinter/linter-eslint/pull/915 this was missed. Luckily the problem it was meant to solve was accepted as a bug against ESLint itself and fixed there which is why this wasn't caught earlier.

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -160,7 +160,6 @@ export function getCLIEngineOptions(type, config, rules, filePath, fileDir, give
   const cliEngineConfig = {
     rules,
     ignore: !config.disableEslintIgnore,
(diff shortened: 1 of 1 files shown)


IanVS:
@Arcanemagus do you have a link to the fix within ESLint?

IanVS:
Found the link you posted on the other PR: https://github.com/eslint/eslint/pull/6632.

Arcanemagus:
Sorry, only posted it there since that's where the problem was introduced.

PR #1070, opened 2018-01-10
Don't return promise until makeFixes completes

skylize:
The makeFixes function is now properly determinent, as a consumer cannot
receive a resolved promise before all lines of makeFixes have executed.
This might (appears to so far) resolve problems with flaky specs.

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -84,23 +84,28 @@ async function getNotification(expectedMessage) {
 }
 
 async function makeFixes(textEditor) {
-  const editorReloaded = new Promise((resolve) => {
-    // Subscribe to the file reload event
-    const reloadSubscription = textEditor.getBuffer().onDidReload(() => {
-      reloadSubscription.dispose()
+  const editorReloadPromise = new Promise((resolve) => {
+    // Subscribe to file reload events
+    const editorReloadSubscription = textEditor.getBuffer().onDidReload(() => {
+      editorReloadSubscription.dispose()
       resolve()
     })
   })
 
-  // Now that subscription is active, send off a fix request
-  atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
   const expectedMessage = 'Linter-ESLint: Fix complete.'
-  const notification = await getNotification(expectedMessage)
+  // Subscribe to notification events
+  const notificationPromise = getNotification(expectedMessage)
 
+  // Subscriptions now active for Editor Reload and Message Notification
+  // Send off a fix request.
+  atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
+
+  const notification = await notificationPromise
   expect(notification.getMessage()).toBe(expectedMessage)
   expect(notification.getType()).toBe('success')
 
-  return editorReloaded
+  // After editor reloads, it should be safe for consuming test to resume.
+  return editorReloadPromise
 }
 
 describe('The eslint provider for Linter', () => {
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -218,9 +218,10 @@ describe('The eslint provider for Linter', () => {
 
     it('will not give warnings when autofixing the file', async () => {
       const editor = await atom.workspace.open(paths.ignored)
-      atom.commands.dispatch(atom.views.getView(editor), 'linter-eslint:fix-file')
       const expectedMessage = 'Linter-ESLint: Fix complete.'
-      const notification = await getNotification(expectedMessage)
+      const notificationPromise = getNotification(expectedMessage)
+      atom.commands.dispatch(atom.views.getView(editor), 'linter-eslint:fix-file')
+      const notification = await notificationPromise
 
       expect(notification.getMessage()).toBe(expectedMessage)
     })
@@ -474,16 +475,18 @@ describe('The eslint provider for Linter', () => {
     })
 
     it('shows an info notification', async () => {
+      const notificationPromise = getNotification(expectedMessage)
       atom.commands.dispatch(atom.views.getView(editor), 'linter-eslint:debug')
-      const notification = await getNotification(expectedMessage)
+      const notification = await notificationPromise
 
       expect(notification.getMessage()).toBe(expectedMessage)
       expect(notification.getType()).toEqual('info')
     })
 
     it('includes debugging information in the details', async () => {
+      const notificationPromise = getNotification(expectedMessage)
       atom.commands.dispatch(atom.views.getView(editor), 'linter-eslint:debug')
-      const notification = await getNotification(expectedMessage)
+      const notification = await notificationPromise
       const detail = notification.getDetail()
 
       expect(detail.includes(`Atom version: ${atom.getVersion()}`)).toBe(true)
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -84,23 +84,23 @@ async function getNotification(expectedMessage) {
 }
 
 async function makeFixes(textEditor) {
-  return new Promise(async (resolve) => {
+  const editorReloaded = new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


spec/linter-eslint-spec.js
```diff
@@ -84,23 +84,23 @@ async function getNotification(expectedMessage) {
...
 
 async function makeFixes(textEditor) {
-  return new Promise(async (resolve) => {
+  const editorReloaded = new Promise((resolve) => {
     // Subscribe to the file reload event
-    const editorReloadSub = textEditor.getBuffer().onDidReload(async () => {
-      editorReloadSub.dispose()
-      // File has been reloaded in Atom, notification checking will happen
-      // async either way, but should already be finished at this point
+    const reloadSubscription = textEditor.getBuffer().onDidReload(() => {
```
Arcanemagus: I'd prefer the old variable name since it's more descriptive of what it actually is.

spec/linter-eslint-spec.js
```diff
@@ -84,23 +84,23 @@ async function getNotification(expectedMessage) {
...
+      reloadSubscription.dispose()
       resolve()
     })
+  })
 
-    // Now that all the required subscriptions are active, send off a fix request
-    atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
-    const expectedMessage = 'Linter-ESLint: Fix complete.'
-    const notification = await getNotification(expectedMessage)
+  // Now that subscription is active, send off a fix request
```
Arcanemagus: How about this?
```js
// Editor reload subscription is now active, send ...
```

spec/linter-eslint-spec.js
skylize: I changed it because I felt like `Sub' was a meaningless abbreviation until I dug into the code to work it out. But I can also see why you would want it to specify what is reloading. It's a little on the long side, but how about `editorReloadSubscription`?

spec/linter-eslint-spec.js
skylize: Yeah, that's fine for me. I only even messed with that comment to make it singular. The old comment falsely implied multiple subscriptions, while there is only one subscription per call to `makeFixes`.

spec/linter-eslint-spec.js
Arcanemagus: `editorReloadSubscription` works for me.

spec/linter-eslint-spec.js
Arcanemagus: There are actually two, `getNotification` creates a subscription on new Notifications in order to grab the one we are waiting for. _Technically_ there is a race condition between dispatching the command and the Notification subscription being added, but the few _us_ that takes is far outweighed by the time it takes to call out to ESLint to make the fixes.

skylize: Hmmm. I didn't catch that.

The race between `getNotifications` and `textEditor.getBuffer().onDidReload` proved quite enough to make tests flaky. Based on the comments, I assume the previous developer believed that it was "technically" a race condition.

So I guess we need to 
```
subscribe to reload
subscribe to notifications
dispatch commands
await notifications
return reload promise
```
?

Arcanemagus: > The race between `getNotifications` and `textEditor.getBuffer().onDidReload` proved quite enough to make tests flaky.

Do you have any basis for this? Although the changes in this PR are more correct, I haven't seen any issues with this in a while.

The only falkiness I know of was actually due to `onDidReload` being completely broken in the default Jasmine environment. (["previous" developer](https://github.com/AtomLinter/linter-eslint/commit/a9269577da08c4de93a341146221b6c1dbcf5ce9#diff-15b3e4c0d78ceab6572aceecc5b4bd6cR204) 👋)

This is fixed in `jasmine-fix` [here](https://github.com/steelbrain/jasmine-fix/blob/c7bc55cc9d6de1f4493f3f7a1930c60b80c00557/index.js#L15-L23).

----

I couldn't think of a good way to have a nice `async` function to get a specific notification that was usable in multiple spots due to the callback nature of the `onDid____` functions.

skylize: > The only falkiness I know of was actually due to onDidReload being completely broken in the default Jasmine environment. 

Since "flakiness" is defined by it's inconsistent nature, there's no real way to know whether it was fixed by that patch. But since it's related to promises and in particular `onDidReload`... maybe? 

> Do you have any basis for this? Although the changes in this PR are more correct, I haven't seen any issues with this in a while.

This fix was based on which specs [reported failure](https://github.com/AtomLinter/linter-eslint/pull/1009#issuecomment-329997867) in randomly failing CircleCI builds. This was about a week before your jasmine fix, and `makeFix` does call `onDidReload`. So honestly I have no idea now.

Maybe I'm just being a perfectionist? I don't see why you would want to even consider allowing race conditions in specs. Are the _potentially_ saved microseconds during testing really worth having indeterminate tests?

It's all running in a single thread, so everything is at best concurrent, and never parallel. If you actually even manage to save some time (which you certainly might fail to accomplish, due to the nature v8's parsing and compiling optimizations), the potential savings is quite negligible.

Arcanemagus: > This fix was based on which specs [reported failure](https://github.com/AtomLinter/linter-eslint/pull/1009#issuecomment-329997867) in randomly failing CircleCI builds. This was about a week before your jasmine fix, and `makeFix` does call `onDidReload`. So honestly I have no idea now.

Ah, if I remember correctly that was actually what triggered my deep investigation into what in the world was going on with the specs and why they were so flaky 😉.

> Maybe I'm just being a perfectionist? I don't see why you would want to even consider allowing race conditions in specs.

Ideally we wouldn't at all, that's not being a perfectionist, just writing good code.

> Are the _potentially_ saved microseconds during testing really worth having indeterminate tests?

I'm not sure what this has to do with performance?

The current design of the `getNotifications` function is only to get the code that is used in many places all into a single spot. The issue with doing it correctly is that each place this function is currently being called would have to be changed to setup the observer and wrap the callback itself. If you have an alternative design for this I'd love to hear it! I don't like relying on this race condition either, no matter how impossible it _should_ be to trigger it.

skylize: Proof that current master branch still has flaky tests.

image

skylize:
Ready for another review.

Re-posting this proof that master branch still has flaky tests, since it seems to be hidden now inside "Show outdated" foldout and I don't know if anyone has seen it.

image

Arcanemagus:
Neither of those specs have anything to do with `makeFixes` (or `getNotification`), so that's something else entirely 😮.

skylize:
> ... have anything to do with makeFixes (or getNotification), ...
Lol. Good point. I guess we need to look there next.

spec/linter-eslint-spec.js
```diff
@@ -84,23 +84,28 @@ async function getNotification(expectedMessage) {
...
+  const notificationPromise = getNotification(expectedMessage)
 
-    expect(notification.getMessage()).toBe(expectedMessage)
-    expect(notification.getType()).toBe('success')
-  })
+  // Subscriptions now active for Editor Reload and Message Notification
+  // Send off a fix request.
+  atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
+
+  const notification = await notificationPromise
```
Arcanemagus: Wow, gotta love when solutions like this are staring you right in the face. Would you mind either submitting a PR fixing the other uses of `getNotification`, or just fixing them here?

skylize: Yeah. I'm trying to start looking harder at the actual application code. But if you think this fix is likely to be applicable in other places in the specs, I'll definitely try to hunt them down.

skylize: This one seems to be good to go as-is and affects multiple tests. So let's commit it, and I'll keep digging for more of the same.

skylize:
@Arcanemagus  I went ahead and made the same getNotification fix for all the tests.

Issue #1059, opened 2017-11-28
linter-eslint stalls on one project in Atom


codequistador:
## Issue Type
Bug

## Issue Description
One of my projects is having trouble using eslint. When I open this project and open a .js file the little icon on the bottom of Atom pulses and is yellow. It never changes and I am unable to get an output from running `Linter Eslint: Debug`. I can run the command in my other project ( see output below ). In the troubled project, the `.eslintrc` file is in the root of the project directory, much like other projects that are working.

Thanks for helping!

## Bug Checklist
- Restart Atom [check]
- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not [check, confirmed]
- Paste the output of the `Linter Eslint: Debug` command from the Command Palette below [check]

```
Atom version: 1.22.1
linter-eslint version: 8.4.0
ESLint version: 4.11.0
Hours since last Atom restart: 0
Platform: darwin
Using local project ESLint from: /Users/jdaining/Sites/inst/bridge-talent-management/bridge-talent-web/ui/node_modules/eslint
Current file's scopes: [
  "source.js.jsx"
]
linter-eslint configuration: {
  "lintHtmlFiles": false,
  "useGlobalEslint": false,
  "showRuleIdInMessage": true,
  "disableWhenNoEslintConfig": true,
  "eslintrcPath": "",
  "globalNodePath": "",
  "advancedLocalNodeModules": "",
  "eslintRulesDirs": [],
  "disableEslintIgnore": false,
  "disableFSCache": false,
  "fixOnSave": false,
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic"
  ],
  "rulesToSilenceWhileTyping": [],
  "rulesToDisableWhileFixing": [],
  "ignoreFixableRulesWhileTyping": false
}
```

Arcanemagus:
Are you able to share this project?

codequistador:
Unfortunately no. I can provide more details though.

codequistador:
I figured out a way around it, but it is by no means a solution! If I switch from Project A (trouble project) to a different project (Project B) in the same window, quit Atom, restart Atom (eslint now working on Project B) and then switch back to Project A, eslint works on Project A. If I close Atom and then open it again, eslint is no longer working on Project A. So strange!

Arcanemagus:
[...]

Issue #930, opened 2017-06-02
Uncaught Error: spawn ENOMEM


zdiehlio:
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.17.2 x64
**Electron**: 1.3.15
**OS**: Ubuntu 16.04.2
**Thrown From**: [linter-eslint](https://github.com/AtomLinter/linter-eslint) package 8.2.1

### Stack Trace

Uncaught Error: spawn ENOMEM

```
At internal/child_process.js:313

Error: spawn ENOMEM
    at exports._errnoException (util.js:1026:11)
    at ChildProcess.spawn (internal/child_process.js:313:11)
    at exports.spawn (child_process.js:392:9)
    at Object.exports.fork (child_process.js:70:10)
    at new Task (/opt/atom/resources/app/src/task.js:43:1)
    at initializeESLintWorker (/packages/linter-eslint/main.js:136:21)
```

### Commands

```
```

### Non-Core Packages

```
busy-signal 1.4.1 
emmet 2.4.3 
intentions 1.1.2 
linter 2.0.0 
linter-eslint 8.2.1 
linter-ui-default 1.6.0 
```

skylize:
Closing as duplicate of #927. Let us know if you have any further info that shows this to be a different isssue.

Issue #927, opened 2017-05-26
Linter crashes after adding too many event listeners, channel closed. (8.2.1)


aforty:
## Issue Type
Bug

## Issue Description
Linter crashes and cannot recover.

## Bug Checklist
- Restart Atom
- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- Paste the output of the `Linter Eslint: Debug` command from the Command Palette below

I cannot get the output of `Linter Eslint: Debug` in the console it'll complai about EventEmitter memory leak.

Here is the output of just using linter on my project, it works for a few minutes but seems that once the eventemitter adds too many listeners it crashes after which linter cannot connect to the process and prints `channel closed`.

Using Atom 1.17.2, linter 2.1.4, linter-eslint 8.2.1.

![screen shot 2017-05-26 at 12 53 27](https://cloud.githubusercontent.com/assets/523683/26504497/cea8d9ee-4212-11e7-947d-77e0aa1add4f.png)

kalimantos:
i suggest you to downgrade to 7.3.2 using `apm install linter-eslint@7.3.2` waiting for a fix for this issue

scholtzm:
Getting loads of `EventEmitter memory leak detected` errors with `linter-eslint 8.2.1`. For now I have downgraded back to `linter-eslint@8.1.7` and seems to be OK.

not-an-aardvark:
Is there a stack trace for the `memory leak detected` warning?

IanVS:
What type and version of OS are you all using?  (Looking for any possible common links)

craig-bishell:
I'm also getting this exact error on Windows 10 (Home), Version 1607, OS Build 14393.1198

Have reverted linter-eslint to 7.3.2 (as per previous issue) and that seems to resolve the issue.

IanVS:
Does turning on/off the `Fix errors on save` option have any effect?

craig-bishell:
No, I do not have it enabled by default, and turning it on (and subsequently off again) has no effect.
The only settings I have enabled are "Disable when no ESLint config is found" and "Show Rule ID in messages" - which I believe are the defaults.

joshmeads:
I was experiencing this issue, however, I updated to the current version again yesterday after a few weeks on 8.1.7 and everything seems to be fine again. Appears this isn't the case for everyone though.

IanVS:
[...]

Issue #925, opened 2017-05-25
8.2.0 -- Atom-Beta 1.18 High memory usage


philipchmalts:
## Issue Type
Bug

## Issue Description

Since updating to 8.20 in atom beta 1.18 - beta1 linter-eslint-helper eats up a ton of memory and basically gets stuck linting files sometimes taking over 30s per file. It was working well before updating to the newest version. Has anyone else encountered this issue?

![screen shot 2017-05-25 at 4 49 20 pm](https://cloud.githubusercontent.com/assets/14876812/26470190/d14bdd20-416a-11e7-9b5c-ddf88564f731.png)

## Bug Checklist
- Restart Atom
- Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- Paste the output of the `Linter Eslint: Debug` command from the Command Palette below

```
Linter Eslint: Debug output here

Atom version: 1.18.0-beta1
linter-eslint version: 8.2.1
ESLint version: 3.19.0
Hours since last Atom restart: 0.3
Platform: darwin
Using local project ESLint from: /Users/philipchmalts/Documents/Development/if-client/node_modules/eslint
Current file's scopes: [
  "source.js.jsx",
  "meta.function.js",
  "punctuation.terminator.statement.js"
]
linter-eslint configuration: {
  "disableWhenNoEslintConfig": false,
  "fixOnSave": true,
  "lintHtmlFiles": false,
  "useGlobalEslint": false,
  "showRuleIdInMessage": true,
  "eslintrcPath": "",
  "globalNodePath": "",
  "advancedLocalNodeModules": "",
  "eslintRulesDir": "",
  "disableEslintIgnore": false,
  "disableFSCache": false,
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic"
  ],
  "rulesToSilenceWhileTyping": [],
  "rulesToDisableWhileFixing": []
}
```

cbranch101:
This is pretty much rendering this plugin unusable for me

nielsvanmidden:
Same. Anyone has a solution?

IanVS:
I assume that downgrading back to linter-eslint version `8.1.7` 

I'm not sure why some users are seeing this and others aren't.  If there is any other information you can share about your configurations or setup, that would be helpful.  Does it happen for you on all projects, or only certain ones?  Also, does it still happen if you disable `fixOnSave`?

[...]

PR #1073, opened 2018-01-12
Handle worker crashes

IanVS:
Refs #925 #927 #930 #1059

This is an attempt to mitigate some of the problems being experienced in the referenced issues, whereby linter-eslint crashes/hangs with a high memory usage.

I was unable to determine the exact cause of workers dying, but this PR does a few things to try to minimize the crashes and memory usage, as well as allowing linter-eslint to recover from a crashed worker.  In one of my projects, I was able to reliably reproduce the crashes, usually within a few seconds of opening a file.  Using this branch, linter-eslint has now become usable for me again, albeit during a crash it still takes a few seconds to recover and start working again, so I don't consider this a complete solution.

I think the main challenge we are facing is that the Task api from Atom is not intended to be used as a long-running task, which is what we need because starting up a lint job can take a second or two.  Maybe we can find another, more reliable approach in the future, but hopefully this PR will at least make `linter-eslint` usable again for those folks currently stuck on the old version.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -43,8 +43,10 @@ export async function sendJob(worker, config) {
   config.emitKey = cryptoRandomString(10)
 
   return new Promise((resolve, reject) => {
-    const errSub = worker.on('task:error', (...err) => {
-      const [msg, stack] = err
+    // All worker errors are caught and re-emitted along with their associated
+    // emitKey, so that we do not create multiple listeners for the same
+    // 'task:error' event
+    const errSub = worker.on(`workerError:${config.emitKey}`, ({ msg, stack }) => {
       // Re-throw errors from the task
       const error = new Error(msg)
       // Set the stack to the one given to us by the worker
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -49,6 +49,9 @@ export async function sendJob(worker, config) {
       const error = new Error(msg)
       // Set the stack to the one given to us by the worker
       error.stack = stack
+      errSub.dispose()
+      // eslint-disable-next-line no-use-before-define
+      responseSub.dispose()
       reject(error)
     })
     const responseSub = worker.on(config.emitKey, (data) => {
@@ -60,6 +63,8 @@ export async function sendJob(worker, config) {
     try {
       worker.send(config)
     } catch (e) {
+      errSub.dispose()
+      responseSub.dispose()
       console.error(e)
     }
   })
--- a/src/main.js
+++ b/src/main.js
@@ -177,12 +177,26 @@ module.exports = {
         }
       }]
     }))
+    this.initializeWorker()
+  },
 
-    const initializeESLintWorker = () => {
-      this.worker = new Task(require.resolve('./worker.js'))
-    }
-    // Initialize the worker during an idle time
-    window.requestIdleCallback(initializeESLintWorker)
+  async initializeWorker() {
+    return new Promise((resolve) => {
+      let callbackId
+      if (this.worker !== null) {
+        this.worker.terminate()
+        this.worker = null
+      }
+
+      const initializeESLintWorker = () => {
+        this.worker = new Task(require.resolve('./worker.js'))
+        idleCallbacks.delete(callbackId)
+        resolve()
+      }
+      // Initialize the worker during an idle time
+      callbackId = window.requestIdleCallback(initializeESLintWorker)
+      idleCallbacks.add(callbackId)
+    })
   },
 
   deactivate() {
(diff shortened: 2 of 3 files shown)


src/main.js
```diff
@@ -195,6 +195,21 @@ module.exports = {
...
+      if (this.worker !== null) {
+        this.worker.terminate()
+        this.worker = null
+      }
+      const initializeESLintWorker = () => {
+        this.worker = new Task(require.resolve('./worker.js'))
+        resolve()
+      }
+      // Initialize the worker during an idle time
+      window.requestIdleCallback(initializeESLintWorker)
```
Arcanemagus: This should be added to the `idelCallbacks` `Set` (The call up on L185 also should be added). Otherwise we risk initializing a worker when the package is deactivated.

src/main.js
```diff
@@ -195,6 +195,21 @@ module.exports = {
     this.subscriptions.dispose()
   },
 
+  async restartESLintWorker() {
+    return new Promise((resolve) => {
+      if (this.worker !== null) {
+        this.worker.terminate()
+        this.worker = null
+      }
+      const initializeESLintWorker = () => {
```
Arcanemagus: Since this code is in multiple places it should be moved to a separate function of its own.

src/main.js
```diff
@@ -323,6 +344,13 @@ module.exports = {
       await waitOnIdle()
     }
 
+    // Sometimes the worker dies and becomes disconnected
```
Arcanemagus: Since this is duplicated maybe this comment as well as the check should be moved into the called function, and it renamed to something like `checkESLintWorker`?

src/worker.js
```diff
@@ -83,42 +83,48 @@ function fixJob({ cliEngineOptions, contents, eslint, filePath }) {
...
-    }
-
-    const relativeFilePath = Helpers.getRelativePath(fileDir, filePath, config, projectPath)
-
-    const cliEngineOptions = Helpers
-      .getCLIEngineOptions(type, config, rules, relativeFilePath, fileDir, configPath)
+    // We catch all worker errors so that we can create a separate error emitter
+    // for each emitKey, rather than adding multiple listeners for `task:error`
+    try {
+      const {
```
Arcanemagus: Minor nit, but I'd leave this out of the `try/catch` and just use `emitKey` in the `catch` block.

src/main.js
IanVS: I went back and forth on it, and in the end decided against passing yet another argument to sendJob, but can do so if you'd prefer that.

Arcanemagus: I actually rather like the idea of just sticking this logic in `sendJob`, that way the logic for maintaining the worker stays there leaving the callers to just worry about how the job is used.

I'm not sure why moving the logic there would require an additional argument though?

src/helpers.js
```diff
@@ -43,12 +43,17 @@ export async function sendJob(worker, config) {
   config.emitKey = cryptoRandomString(10)
 
   return new Promise((resolve, reject) => {
```
skylize: Why do we have a promise that can never resolve?

skylize: Nevermind, resolve was hidden in folded code. 🙈

Arcanemagus: That's what the `reject` on L57, and the `resolve` on L62 are for?

src/main.js
IanVS: Is there a way to kill and restart a worker simply from a reference to that worker?  You can see that my `restartESLintWorker()` method sets `this.worker`, which is then passed to `sendJob`.  So I assume we would need to send a reference to `this.restartESLintWorker` to `sendJob` as well.  Or am I missing another approach that you have in mind?

Arcanemagus: I was thinking this bit:
```js
if (this.worker && !this.worker.childProcess.connected) {
  await this.initializeWorker()
}
```

As well as the code in `initializeWorker` could be moved into `sendJob` (or called from there).

IanVS: Hm, I guess I'm not seeing a clear vision of what you have in mind.  Do you want to take a crack at a commit on this PR?

skylize: I'm not sure if this is at all an improvement. A lot of side effects and misdirection, but I'll share the idea anyways.

````
const advancedWorker = worker => ({
  initialize: () => {
    this.initializeEslintWorker();
    return this.worker
  },
  restart: {
    this.restartEslintWorker,
    return this.worker
  }
  task: this.worker
})

function sendJob( worker, config) {
  if (!worker.task || !this.worker.childProcess.connected) {
    worker.task = worker.initialize()
  }
  // ....
}
```

src/main.js
```diff
@@ -177,21 +162,12 @@ module.exports = {
...
   },
 
   deactivate() {
-    if (this.worker !== null) {
-      this.worker.terminate()
-      this.worker = null
-    }
     idleCallbacks.forEach(callbackID => window.cancelIdleCallback(callbackID))
     idleCallbacks.clear()
+    helpers.killWorker()
```
Arcanemagus: Note that this is _after_ the cancellation of the idle callbacks to prevent the idle callback that requests a worker start from firing off before it would be able to kill it.

src/helpers.js
```diff
@@ -83,7 +103,7 @@ function validatePoint(textBuffer, line, col) {
   }
 }
 
-export async function getDebugInfo(worker) {
+export async function getDebugInfo() {
```
skylize: I like this a lot. Gets a ton of the worker logic and state management of the worker out the the `main` object, greatly simplifying things. Definitely going the right direction IMHO.

PR #1074, opened 2018-01-13
Remove useless "if" wrappers around requires

skylize:
Node's `require` stores imports in memory. If you require the same module
multiple times, it will reuse the first import each time. Using late requires is not ideal because `require` is blocking when it does need to load a file, but if we are going to use it, we can at least use it as designed.

Patch removes if checks around requires and replaces global let
variables with const variables local to the function that uses the
imports.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -4,12 +4,8 @@
 import { CompositeDisposable, Task } from 'atom'
 
 // Dependencies
-// NOTE: We are not directly requiring these in order to reduce the time it
-// takes to require this file as that causes delays in Atom loading this package
-let path
-let helpers
-let workerHelpers
-let isConfigAtHomeRoot
+// NOTE: Some dependencies are not listed here because they are not required
+// until expected first use to reduce initial loading time of this package.
 
 // Configuration
 const scopes = []
@@ -113,9 +109,7 @@ module.exports = {
 
     this.subscriptions.add(atom.commands.add('atom-text-editor', {
       'linter-eslint:debug': async () => {
-        if (!helpers) {
-          helpers = require('./helpers')
-        }
+        const helpers = require('./helpers')
         if (!this.worker) {
           await waitOnIdle()
         }
@@ -202,6 +196,8 @@ module.exports = {
       scope: 'file',
       lintsOnChange: true,
       lint: async (textEditor) => {
+        const helpers = require('./helpers')
+
         if (!atom.workspace.isTextEditor(textEditor)) {
           // If we somehow get fed an invalid TextEditor just immediately return
           return null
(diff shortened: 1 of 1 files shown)


skylize:
Maybe I am misunderstanding the implementation of `require`? From what I understand, node's `require` saves modules in a cache in memory and pulls it from the cache when called again for the same module (I have no idea about other commonjs implementations, but I believe this is the case in node). I would not expect it to hit the file system again after the first call to a given module.

If I am wrong about this, I would be happy to use our own caching instead. The current setup is quite gross, with these global let variables.

Arcanemagus:
Run a CPU profile of the code and you can see that it still takes some time in `fs` lookups, this is I think mostly due to Atom's customizations of `require` (to make it automatically use the transpiled version of code).

skylize:
> Also there is a huge difference between globally scoped variables and file scoped variables.

That is a very good point, but when you are looking at large files, the difference starts to decrease. A file scoped variable is a whole lot more "global" in a 350 line file like `main.js` with a humongous object that borders on being an OOP Class with long imperative methods than it is in a 10-20 line file with a few simple and pure declarative functions.

> Run a CPU profile of the code and you can see that it still takes some time in fs lookups

I am _definitely_ not a Profiling expert, but in my attempt to profile, one spec stood out pretty clearly to me, the total execution time of all the specs. 

After numerous attempts, both master and this patch showed a range of  52-61 seconds to complete the full spec suite on my system. Master branch had a  tendency toward the higher end of that range, while the patch had a tendency toward the lower end of that range. 

The patch actually spent noticeably _less_ time in `fs`, but spent a whole lot more tie in `TextBuffer`. All-in-all, my interpretation is there is no reliably measureable difference between the two, but if I was forced to place a bet, I would put my money on the patch. 

If you feel like you can do a better job of interpreting cpu profiles, I am all ears.

As it stands _my opinion_ is that with no obvious performance impact, we should aim for  maintainability, readability, and refactorability. I think all three of those are improved by eliminating some unnecessary mutable (file-scoped) global variables and replacing several 3-line if-wrapped  requires with simple idiomatic 1-line requires.

And if you still disagree for some other reason than performance, I'm still more than happy to work on a simple late-loader or module cache instead.

IanVS:
I would also want to spend a little time testing this out and profiling the impacts.  Honestly, I think I'd rather we went to the `loadDeps` strategy that @Arcanemagus mentioned.  Would you be opposed to that approach, @skylize?

skylize:
@IanVS I should have cross linked, instead of linking only one way. Please also see PR #1075.

Arcanemagus:
Closing this in favor of #1075 as it is much closer to what I'd like to see here, especially since it sounds like @IanVS is of the same opinion.

PR #1077, opened 2018-01-16
Update atom-babel6-transpiler to v1.1.3

Arcanemagus:
Brings in a few bug fixes to the minimum version allowed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "lint": "eslint ."
   },
   "dependencies": {
-    "atom-babel6-transpiler": "^1.1.1",
+    "atom-babel6-transpiler": "^1.1.3",
     "atom-linter": "^10.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -119,7 +119,7 @@
     "lint": "eslint ."
   },
   "dependencies": {
-    "atom-babel6-transpiler": "^1.1.1",
+    "atom-babel6-transpiler": "^1.1.3",
     "atom-linter": "^10.0.0",
     "atom-package-deps": "^4.6.0",
     "babel-preset-node5": "^12.0.0",
```
skylize: Can we update this one too? We should be transpiling to 7.4.

Arcanemagus: Moving to `babel-preset-env` is something I want to do in a separate PR.

PR #1078, opened 2018-01-16
Switch to babel-preset-env

Arcanemagus:
Switch to using `babel-preset-env` instead of `babel-preset-node5`. This allows our transpilation to [dynamically adapt](https://www.npmjs.com/package/babel-preset-env#targetsnode) to the currently running version of Node.js within Atom instead of being forced down to Node.js v5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     "atom-babel6-transpiler": "^1.1.3",
     "atom-linter": "^10.0.0",
     "atom-package-deps": "^4.6.0",
-    "babel-preset-node5": "^12.0.0",
+    "babel-preset-env": "^1.6.1",
     "consistent-path": "^2.0.1",
     "crypto-random-string": "^1.0.0",
     "eslint": "^4.9.0",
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Note that when Babel 7 releases this is being moved to `@babel/preset-env`, however there is currently no Babel 7 Atom transpiler, I've filed https://github.com/atom/atom-babel6-transpiler/issues/5 to see about either adding support there, or creating a new package for that.

package.json
```diff
@@ -197,7 +197,14 @@
         "setBabelEnv": false,
         "babel": {
           "presets": [
-            "node5"
+            [
+              "env",
+              {
+                "targets": {
+                  "node": "current"
```
skylize: Haha. 'current' right now is 9.4.0. Atom won't be able to read any of it if you transpile to that.

Change this to `"node": "7.4.0".

Arcanemagus: That's not what that actually means. As [the documentation](https://www.npmjs.com/package/babel-preset-env#targetsnode) states `current` means "use the version of Node.js that this is _currently_ running under.

Arcanemagus: Updated the PR description to link there as well so this was a bit clearer 😉.

skylize: Hmm, interesting. I did not realize that. Great call then. That will keep it evergreen. :🤸‍♂️

PR #1084, opened 2018-01-19
Merge all rules to ignore while typing

### LINKED ISSUE #1081: ignoreFixableRulesWhileTyping overrides ignoredRulesWhenModified
## Issue Type
Bug

## Issue Description
If the user has enabled the `ignoreFixableRulesWhileTyping` option _and_ has rules set in `ignoredRulesWhenModified` the rules specified in `ignoredRulesWhenModified` are overridden by the list of rules determined to be fixable.

These two lists should be combined into a single list of rules to ignore.

## Bug Checklist
- [X] Restart Atom
- [X] Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- [X] Paste the output of the `Linter Eslint: Debug` command from the Command Palette below

Arcanemagus:
Merge the list of rules from `ignoreFixableRulesWhileTyping` and `ignoredRulesWhenModified` into a single list that is sent to ESLint.

Fixes #1081.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -61,12 +61,18 @@ let disableWhenNoEslintConfig
 let ignoreFixableRulesWhileTyping
 
 // Internal functions
+/**
+ * Given an Array or iterable containing a list of Rule IDs, return an Object
+ * to be sent to ESLint's configuration that disables those rules.
+ * @param  {[iterable]} ruleIds Iterable containing ruleIds to ignore
+ * @return {Object}             Object containing properties for each rule to ignore
+ */
 const idsToIgnoredRules = ruleIds =>
-  ruleIds.reduce((ids, id) => {
-    // eslint-disable-next-line no-param-reassign
-    ids[id] = 0 // 0 is the severity to turn off a rule
-    return ids
-  }, {})
+  Array.from(ruleIds).reduce(
+    // 0 is the severity to turn off a rule
+    (ids, id) => Object.assign(ids, { [id]: 0 })
+    , {}
+  )
 
 // Worker still hasn't initialized, since the queued idle callbacks are
 // done in order, waiting on a newly queued idle callback will ensure that
@@ -168,7 +174,7 @@ module.exports = {
 
     this.subscriptions.add(atom.config.observe(
       'linter-eslint.rulesToSilenceWhileTyping',
-      (ids) => { ignoredRulesWhenModified = idsToIgnoredRules(ids) }
+      (ids) => { ignoredRulesWhenModified = ids }
     ))
 
     this.subscriptions.add(atom.config.observe(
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -269,12 +275,14 @@ module.exports = {
...
 
         let rules = {}
-        if (textEditor.isModified() && Object.keys(ignoredRulesWhenModified).length > 0) {
-          rules = ignoredRulesWhenModified
-        }
-        if (textEditor.isModified() && ignoreFixableRulesWhileTyping) {
-          // Note that this list will only contain rules after the first lint job
-          rules = idsToIgnoredRules(helpers.rules.getFixableRules())
+        if (textEditor.isModified()) {
+          const ignoredRules = new Set()
```
IanVS: Is there any way to avoid creating a new set and looping through all the `ignoredRulesWhenModified` and fixable rules each time the editor is modified?

Arcanemagus: The fixable rules could change at any time, so there isn't _too_ much point in saving the generated list.

It's certainly possible though.

skylize: This seems a lot cleaner than the `forEach` loop.
```
const ignoredRules = new Set(ignoredRulesWhenModified)
```

How about this for this whole section?
```
const ignoredRules = [
  ...ignoredRulesWhenModified,
  ...(
    ignoreFixableRulesWhileTyping
      ? helpers.rules.getFixableRules()
      :  []
  )
]
const uniqueIgnoredRules = new Set(ignoredRules)` 
rules = idsToIgnoredRules(uniqueIgnoredRules)
```

IanVS: What if we use `if (textEditor.isModified() && ignoreFixableRulesWhileTyping)` at least?

Arcanemagus: That's essentially what's already happening?

```js
if (textEditor.isModified()) {
  // ...
  if (ignoreFixableRulesWhileTyping) {
```

Arcanemagus: > `const ignoredRules = new Set(ignoredRulesWhenModified)`

Definitely, changed.

> How about this for this whole section?

It's clearer to me at least the way it is.

IanVS: By moving that check up top, if `ignoreFixableRulesWhileTyping` is false, we don't need to create a new `ignoredRules` set and move all of the `ignoredRulesWhenModified` into it.

skylize: Once we have this in a data-store, we can use `reselect` to memoize a selector on this data.

skylize:
> By moving that check up top, if `ignoreFixableRulesWhileTyping` is false, we don't need to create a new ignoredRules set and move all of the `ignoredRulesWhenModified` into it.

Want to reopen this statement outside the "outdated" comment folds. I'm actually a little unclear if this would work as suggested, but it sounds good if it does.

src/main.js
Arcanemagus: Handling of both `ignoreFixableRulesWhileTyping` and `ignoredRulesWhenModified` is gated behind `textEditor.isModified()`, so in this I just moved them both under a single check since there is no reason to check that twice.

Is this what you mean?
```js
if (textEditor.isModified()) {
  rules = idsToIgnoredRules(ignoredRulesWhenModified)
  if (ignoreFixableRulesWhileTyping) {
    // Note that this list will only contain rules after the first lint job
    rules = Object.assign(rules, idsToIgnoredRules(helpers.rules.getFixableRules())
  }
}
```

src/main.js
```diff
@@ -269,12 +275,13 @@ module.exports = {
...
-        if (textEditor.isModified() && ignoreFixableRulesWhileTyping) {
-          // Note that this list will only contain rules after the first lint job
-          rules = idsToIgnoredRules(helpers.rules.getFixableRules())
+        if (textEditor.isModified()) {
+          const ignoredRules = new Set(ignoredRulesWhenModified)
+          if (ignoreFixableRulesWhileTyping) {
+            // Note that this list will only contain rules after the first lint job
+            helpers.rules.getFixableRules().forEach(ruleId => ignoredRules.add(ruleId))
+          }
+          rules = idsToIgnoredRules(ignoredRules)
```
IanVS: How about:

Arcanemagus: Sure, I'm fine with that.

PR #1087, opened 2018-01-22
Update: Store rules w/ redux, and diff rules

skylize:
Adds a redux data-store, with reducer and actions for storing `rules` in
the store, and selectors for retrieving computed data. Adds diffing to
rules to reduce overhead of passing between threads. Includes tests for
new code, a few generic utility functions, and directory structure
preimptively laid out for dealing with more pieces of state in the
store.


src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
...
+import { deepProp } from '../../f-utils'
+import * as rules from './rules'
+import diffDict from '../../f-utils/deep-diff-dict'
+
+
+// Grab any arbitrary property from the state tree with
+// simple string representation { deepKey: 'some.deep.prop' }
+// or null if does not exist.
+//
+export const selectDeepProp = createSelector(
```
Arcanemagus: 🔥 This function is never actually used, it only exists here and in the specs for it.

src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
+'use babel'
+
+import { createSelector } from 'reselect'
+import ruleDocURIs from 'eslint-rule-documentation'
+import r from '../../f-utils/mini-ramda'
+import { deepProp } from '../../f-utils'
```
Arcanemagus: 🔥 Only used in `selectDeepProp`, which isn't used.

src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
+'use babel'
+
+import { createSelector } from 'reselect'
+import ruleDocURIs from 'eslint-rule-documentation'
+import r from '../../f-utils/mini-ramda'
```
Arcanemagus: 🔥 Only used in `selectDeepProp`, which isn't used.

src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
...
+import { deepProp } from '../../f-utils'
+import * as rules from './rules'
+import diffDict from '../../f-utils/deep-diff-dict'
```
Arcanemagus: This is a similar story to `ruleDocURIs`, but while both of these _never_ change, this at least might have the possibility of changing in the future so I'm not quite as strongly feeling that this _needs_ to be moved into `rules.js`, I'd definitely _prefer_ it though as it makes it clearer where this is actually needed.

src/main.js
```diff
@@ -2,6 +2,10 @@
 
 // eslint-disable-next-line import/no-extraneous-dependencies, import/extensions
 import { CompositeDisposable } from 'atom'
+import store from './store/main'
+import { getFixableAsIgnored } from './store/selectors'
+// eslint-disable-next-line import/prefer-default-export
+export { store }
```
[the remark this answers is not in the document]
skylize: If you are referring to the fact that I'm both importing and exporting `store`, it because it is essentially an instance, so I need to send a reference to that same instance to the helpers. If we were merging this, I could probably make it more clear by _not_ calling `createStore` automatically on the export default this is importing from.  So then I would need to call `createStore()` here and it would be more clear that it was an instance. With a little work and refactoring, the eventual goal would have been that any functions in helpers that need to know about the state or call dispatch would be passed the state or the store as a parameter.

src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
+'use babel'
+
+import { createSelector } from 'reselect'
+import ruleDocURIs from 'eslint-rule-documentation'
```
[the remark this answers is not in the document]
skylize: I don't quite agree that it "should", but I can go along with "acceptable".

src/store/selectors/index.js
skylize: hmm, maybe some cleanup needed

src/store/selectors/index.js
skylize: Yeah, it's not used now. It was just something tiny I threw together while I was in the right mindset, so when the need arose, which is inevitable if we use this, we would already have it ready.

src/store/selectors/index.js
skylize: same, response. If we move the other one, this should go too though.

src/store/selectors/index.js
```diff
@@ -0,0 +1,42 @@
...
+import r from '../../f-utils/mini-ramda'
+import { deepProp } from '../../f-utils'
+import * as rules from './rules'
+import diffDict from '../../f-utils/deep-diff-dict'
+
+
+// Grab any arbitrary property from the state tree with
+// simple string representation { deepKey: 'some.deep.prop' }
+// or null if does not exist.
+//
```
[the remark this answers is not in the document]
skylize: Yeah, me too. But I also like white space for readability. That's why there is _not_ an empty line, but there _is_ a line with an empty comment. That empty comment attaches the comment text above it to the code while still leaving some nice whitespace.

PR #1088, opened 2018-01-22
Update: Extract some validation to separate files

skylize:
Move validation of TextBuffer points and editor scopes to a separate
files in  folder dedicated to validation logic..

### DIFF
--- a/spec/validate/editor-spec.js
+++ b/spec/validate/editor-spec.js
@@ -1,0 +1,59 @@
+'use babel'
+
+import { isValidPoint, hasValidScope } from '../../src/validate/editor'
+import makeSpy from '../make-spy'
+
+// NOTE throwIfInvalidPoint not covered,
+// but it is a simple composition of 2 tested functions.
+
+describe('isValidPoint', () => {
+  it('returns true if clipped point reports being equal to point', () => {
+    const mockPoint = { isEqual: () => true }
+    const spy = makeSpy(mockPoint)
+    const mockTextBuffer = { clipPosition: spy.call }
+    const point = [34, 110]
+
+    const result = isValidPoint(mockTextBuffer, ...point)
+    expect(spy.calledWith[0][0]).toEqual(point)
+    expect(result).toBe(true)
+  })
+
+  it('returns false if clipped point reports not being equal to point', () => {
+    const mockPoint = { isEqual: () => false }
+    const spy = makeSpy(mockPoint)
+    const mockTextBuffer = { clipPosition: spy.call }
+    const point = [12, 14]
+
+    const result = isValidPoint(mockTextBuffer, ...point)
+    expect(spy.calledWith[0][0]).toEqual(point)
+    expect(result).toBe(false)
+  })
+})
+
+describe('hasValidScope', () => {
+  it('returns true if scopes array contains some value in validScopes', () => {
+    const mockEditor = {
+      getCursors: () => [{
+        getScopeDescriptor: () => ({
+          getScopesArray: () => ['valid.scope']
+        })
+      }]
+    }
+    const scopes = ['valid.scope']
+    const result = hasValidScope(mockEditor, scopes)
+    expect(result).toBe(true)
+  })
+
+  it('returns false when scopes array has no values in validScopes', () => {
+    const mockEditor = {
+      getCursors: () => [{
+        getScopeDescriptor: () => ({
+          getScopesArray: () => ['someother.scope']
+        })
+      }]
+    }
+    const scopes = ['invalid.scope']
+    const result = hasValidScope(mockEditor, scopes)
+    expect(result).toBe(false)
+  })
+})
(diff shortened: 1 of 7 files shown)


src/validate/editor.js
```diff
@@ -0,0 +1,20 @@
...
+import throwIfFail from './throw'
+
+export const isValidPoint = (textBuffer, line, column) => {
+  // Clip the given point to a valid one
+  const validPoint = textBuffer.clipPosition([line, column])
+  // Compare to original
+  return validPoint.isEqual([line, column])
+}
+
+export const throwIfInvalidPoint = (tb, line, col) => throwIfFail(
```
Arcanemagus: What's a `tb`? `textBuffer`?

Arcanemagus: For utility functions like this just expand variables like `col` to `column`, they are only typed twice and it makes things perfectly clear.

src/validate/editor.js
```diff
@@ -0,0 +1,20 @@
...
+  (`${line}:${col} isn't a valid point!`),
```
Arcanemagus: Why is this wrapped in `()`?

src/validate/throw.js
```diff
@@ -0,0 +1,9 @@
+'use babel'
+
+const throwIfFail = (msg, passed) => {
+  if (!passed) {
+    throw new Error(msg)
+  }
+}
+
+export default throwIfFail
```
Arcanemagus: Just move the `export default` to the definition above. That way if more things are added later you just need to remove `default`.

src/validate/editor.js
```diff
@@ -0,0 +1,20 @@
...
+  isValidPoint(tb, line, col)
+)
+
+export const isValidScope = (editor, scopes) => editor.getCursors()
```
Arcanemagus: I think I'd say `hasValidScope` is a better name here, the function is checking if the `TextEditor` _has_ a cursor with some scope that matches the list of valid scopes passed in.

Arcanemagus: Let's rename `scopes` to `validScopes` to make it clear what's going on.

src/validate/editor.js
skylize: No problem

src/validate/editor.js
skylize: Artifact from it's old context. Shouldn't be there.

src/validate/editor.js
skylize: Makes sense. Actually I think that little changes makes the entire function noticeably easier to make sense of. 👍 👍

src/validate/throw.js
skylize: I don't think I follow you on "just need to remove `default`. As it is now, it already has a named identifier `throwIfFail`, so if we add more stuff and decide it shouldn't be the default, we can just put `export` in front of the function and move on. To assign it directly to `default`, we must remove the identifier, so if we change it later we must go through the naming process again. Or maybe I'm misunderstanding?

Also I think it is more self documenting this way. If we remove the identifer, we probably need to make the file name significantly more descriptive.

skylize: Sounds great! Only problem is that's not valid Javascript. Lol 

> Error | ESLint | Parsing error: Unexpected token const (Fatal) | 3:16 | src\validate\throw.js

skylize: Also, with current config, if if we try to export it both by name and as default
```
export const throwIfFail = /*... */
export default throwIfFail
```
then ESLint will yell at us later on if we try to import from default.
``` 
import throwIfFail from './validate/throw' // eslint error: import/no-named-as-default
```

Arcanemagus: 😆 Hmmm, I was sure I've used that before. I'm fine with leaving it as is for now.

src/main.js
```diff
@@ -2,7 +2,7 @@
 
 // eslint-disable-next-line import/no-extraneous-dependencies, import/extensions
 import { CompositeDisposable } from 'atom'
-
+import { hasValidScope } from './validate/editor'
```
Arcanemagus: Can you add the blank link after the `import`s back?

spec/validate/editor-spec.js
```diff
@@ -0,0 +1,59 @@
...
+    const spy = makeSpy(mockPoint)
+    const mockTextBuffer = { clipPosition: spy.call }
+    const point = [34, 110]
+
+    const result = isValidPoint(mockTextBuffer, ...point)
+    expect(spy.calledWith[0][0]).toEqual(point)
+    expect(result).toBe(true)
+  })
+
+  it('returns false if clipped point reports being equal to point', () => {
```
Arcanemagus: "reports **not** being equal"

spec/validate/editor-spec.js
```diff
@@ -0,0 +1,59 @@
+'use babel'
+
+import { isValidPoint, hasValidScope } from '../../src/validate/editor'
+import makeSpy from '../make-spy'
+
+// NOTE throwIfInvalidPoint not covered,
+// but it's simple composition of 2 tested functions.
```
Arcanemagus: "it's a simple"

PR #1089, opened 2018-01-22
Docs: Explain Developer Mode for contributors.

skylize:
Document use of `apm link` and `atom -d` for running the working directory as a package. Add some extra detail and fun to other steps. Add spacing to lists for readability.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,47 +4,75 @@
 
 If you would like to contribute enhancements or fixes, please do the following:
 
-1.  Fork the package repository.
-2.  Run `npm install` to setup all dependencies.
-3.  Hack on a separate topic branch created from the latest `master`.
-4.  Check for lint errors with `npm run lint` or use `linter-eslint` within Atom.
-5.  Run the package specs.
-6.  Commit the changes and push the topic branch to your fork.
-7.  Make a pull request to the main repo.
-8.  Welcome to the club!
+1. Fork the package repository and clone your fork to a local folder.
 
-## Guidelines
+1. Run `npm install` in the project to setup all dependencies.
+
+1. Run `apm link --dev` to link your working folder as a development-mode-only package. ([Testing](#testing))
+
+1. Run `git checkout -b my-awesome-branch` to create and checkout a separate branch for your work. ([Tips](#tips))
+
+1. Hack away at the code, and don't forget to have fun. :woman_technologist:
+
+1. Check for lint errors with `npm run lint` or use `linter-eslint` within Atom. ([Guidelines](#code-guidelines))
+
+1. Run `atom --dev path/to/cool-project` to test-drive your improvements. ([Testing](#testing))
+
+1. Run the specs and verify they are all green. ([Testing](#testing))
+
+1. Commit the changes and push the topic branch to your fork on GitHub.
+
+1. Make a pull request to the main repo [here](https://github.com/AtomLinter/linter-eslint/compare).
+
+1. Welcome to the club! :woman_cartwheeling:
+
+# Tips
+
+* The steps above had you checkout a separate branch. This is not strictly necessary. But it is strongly encouraged. The `master` branch might update before your commit is merged. In this case, you will need to rebase those changes into your own commit. Having a separate branch for your work greatly reduces the number of things that can go wrong in that process.
+
+## Code Guidelines
 
 Please note that modifications should follow these coding guidelines:
 
-*   Indent is 2 spaces.
-*   Code should pass the `eslint` linter.
-*   Vertical whitespace helps readability, don’t be afraid to use it!
+* Code should pass linting by `eslint`. In addition to checking for common bugs, the linter is used to enforce style requirements.
+
+* Vertical whitespace helps readability, don’t be afraid to use it!
 
 ## Testing
 
-You can run the specs for this package locally by either running `apm test` from
-the project directory, or from within Atom by going to View -> Developer -> Run
-Package Specs.
+* Running `apm link --dev` will link your working folder to Atom as a development-mode-only package. Atom will load this development version of the package instead of any regularly installed package, allowing you to test your code in a live environment. Use `atom --dev some/project/folder` to start in developer mode.
+
+* There are several ways to start the specs. Any of these will run the full test suite. Make sure the tests pass before submitting a pull request.
+    * Run `atom --test spec` in the command line.
+    * Type `specs` in the Command Pallete and press <kbd>Enter</kbd>.
+    * Click through the menus `View` -> `Developer` -> `Run` -> `Package Specs`
+    * Press the hotkey <kbd>Ctrl</kbd>-<kbd>Y</kbd> or <kbd>⌘</kbd>-<kbd>Alt</kbd>-<kbd>Ctrl</kbd>-<kbd>P</kbd>
+
 
 ## Discussion
-If you have any questions during development you can join the #linter-plugin-dev
+If you have any questions during development you can join the `#linter-plugin-dev`
 channel on [Atom's Slack instance](http://atom-slack.herokuapp.com/).
 
-## Releasing
+## Maintainers
 
-Project members with push access to the repository also have the permissions
+* If given Maintainer status, you should create your branches directly off of the `atom-linter` repo instead of using a fork. This allows other maintainers to easily add additional commits to your branch for an open pull request.
+
+* Project members with push access to the repository also have the permissions
 needed to release a new version.  If there have been changes to the project and
 the team decides it is time for a new release, the process is to:
 
-1.  Update `CHANGELOG.md` with the planned `semver` version number and a short
+    1.  Update `CHANGELOG.md` with the planned `semver` version number and a short
     bulleted list of major changes.  Include pull request numbers if applicable.
-2.  Update `package.json` with the new version number.
-3.  Commit the changed `package.json` and `CHANGELOG.md` to master
-4.  Create a git tag for the new version and push it to GitHub
-  * To create the (signed) tag: `git tag -s -m "vx.y.z" vx.y.z`
-  * To push to GitHub: `git push --follow-tags`
-5.  Publish the package to the Atom package manager
-  * `apm publish --tag vx.y.z`
 
-Thank you for helping out!
+    1.  Update `package.json` with the new version number.
+
+    1.  Commit the changed `package.json` and `CHANGELOG.md` to master
+
+    1.  Create a git tag for the new version and push it to GitHub
+      * To create the (signed) tag: `git tag -s -m "vx.y.z" vx.y.z`
+      * To push to GitHub: `git push --follow-tags`
+
+    1.  Publish the package to the Atom package manager
+      * `apm publish --tag vx.y.z`
+
+Thank you for helping out! :two_hearts:


CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
 
-1.  Fork the package repository.
-2.  Run `npm install` to setup all dependencies.
-3.  Hack on a separate topic branch created from the latest `master`.
-4.  Check for lint errors with `npm run lint` or use `linter-eslint` within Atom.
-5.  Run the package specs.
-6.  Commit the changes and push the topic branch to your fork.
-7.  Make a pull request to the main repo.
-8.  Welcome to the club!
+1. Fork the package repository and clone your fork to local file.
```
Arcanemagus: "to a local folder."

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Run `npm install` in the project to setup all dependencies.
```
Arcanemagus: _Technically_ it should be `apm install`, since then you get the packages installed how Atom will be installing them, but I'm perfectly fine with `npm install` since it will generate a _much_ better tree than the current `npm@3.10.10` bundled in `apm`.

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Run `apm link --dev` to install your working folder as a development-mode-only package in Atom.
```
Arcanemagus: "install" -> "link"
(`install` implies it's a separate copy)

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. While on the `master` branch run `git checkout -b my-super-special-branch` to create and checkout a separate branch for your work. This makes merging your changes a lot easier for the maintainers.
```
Arcanemagus: Maybe add a note about bringing in changes to `master` while you are working on `my-super-special-branch`?

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Hack away at the code, and don't forget to have fun.
+
+1. Check for lint errors with `npm run lint` or use `linter-eslint` within Atom.
+
+1. Run `atom path/to/cool-project --dev` to open Atom in developer mode and test-drive your improvements.
```
Arcanemagus: `atom --dev path/to/cool-project` is the correct form (although Atom works with this form).

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Run the package specs with `Ctrl-Y` or `⌘-Alt-Ctrl-P`, or by typing `specs` in the command pallete.
```
Arcanemagus: ``Ctrl-Y` should be <kbd>Ctrl</kbd>-<kbd>Y</kbd>:
```markdown
<kbd>Ctrl</kbd>-<kbd>Y</kbd>
```

`⌘-Alt-Ctrl-P` should be <kbd>⌘</kbd>-<kbd>Alt</kbd>-<kbd>Ctrl</kbd>-<kbd>P</kbd>:
```markdown
<kbd>⌘</kbd>-<kbd>Alt</kbd>-<kbd>Ctrl</kbd>-<kbd>P</kbd>
```

Arcanemagus: "command pallete" -> "Command Palette"

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Commit the changes and push the topic branch to your fork on GitHub.
+
+1. Make a pull request to the main repo [here](https://github.com/AtomLinter/linter-eslint/pulls).
+
+1. Welcome to the club! 🤸‍♀️
```
Arcanemagus: Please use one of [these emoji](https://www.webpagefx.com/tools/emoji-cheat-sheet/), as GitHub has scripts in place to make them work on _any_ system while the Unicode emoji above only works in certain aspects.

(For example it's broken in source code view, but works in rendered view on my system.)

CONTRIBUTING.md
Arcanemagus: Also, we should probably just enhance the `## Testing` section and just link it here instead.

CONTRIBUTING.md
```diff
@@ -4,22 +4,38 @@
...
+
+1. Check for lint errors with `npm run lint` or use `linter-eslint` within Atom.
+
+1. Run `atom path/to/cool-project --dev` to open Atom in developer mode and test-drive your improvements.
+
+1. Run the package specs with `Ctrl-Y` or `⌘-Alt-Ctrl-P`, or by typing `specs` in the command pallete.
+
+1. Commit the changes and push the topic branch to your fork on GitHub.
+
+1. Make a pull request to the main repo [here](https://github.com/AtomLinter/linter-eslint/pulls).
```
Arcanemagus: Link to `https://github.com/AtomLinter/linter-eslint/compare` if anywhere.

CONTRIBUTING.md
Arcanemagus: Also I notice you've switched to `1.` for all the numbering. I'm not really _against_ that as it does make diffs easier to read, but later in the document another list is back to `n.` numbering. Note that `remark` under the default config will yell at you for this style, I never got around to adding a configuration here for that so I'm not going to enforce either method 😉.

skylize: `npm` was already in the text, I didn't put it there. It's what I used and it worked for me. If you think `apm` is a better instruction to give I can change it.

Not sure how I missed some of the numbering. My intention at least was to change them all. I agree it will give nicer on diffs, and it's also easier to edit. You can move things around, insert or delete, etc. without worrying about renumbering.

Technically I could have just inserted 1 for my new entries and it would have worked itself out, but that seems pretty gross to me. All 1s is a lot less icky.

I don't know anything about `remark`, but I'm happy to change it back if you want, or just finish what I started.

CONTRIBUTING.md
skylize: Hmmm. Yeah I got a bit hung up on that sentence. I was trying to get across that it would have the same effect as an install. 😕

CONTRIBUTING.md
skylize: 👍

CONTRIBUTING.md
skylize: I'm not sure what you're discussing here. Is this about rebasing on master branch to stay up to date?

CONTRIBUTING.md
skylize: Hmm. Yeah I guess you're right. `atom [options] [path...]`. Wow that really _is_ nitpicking. :wink:

CONTRIBUTING.md
skylize: 👍

CONTRIBUTING.md
skylize: 👍

CONTRIBUTING.md
Arcanemagus: I'd say just change them all to `1.`, I slightly prefer that. At some point I'll add `remark` here and we can fix it then if we want to or _enforce_ it then 😉.

CONTRIBUTING.md
Arcanemagus: Yea, I'll see people submit PR's directly from their `master` and then get into _huge_ messes when their PR doesn't get merged before `master` gets changed.

CONTRIBUTING.md
skylize: `:man_carwheeling:` 🤸‍♂️  and `:woman_carthweeling:` 🤸‍♀️  both show in the dropdown when typing into GitHub comments. ??? Do you think it would make a difference if I used one of those instead of the actual emoji generated in the preview tab?

CONTRIBUTING.md
Arcanemagus: Atom didn't used to be that forgiving about which order that was 😉.

CONTRIBUTING.md
Arcanemagus: Ah, exactly! I just assumed it wasn't in the list since you used the raw emoji instead of the text form. GitHub has some script in place where if the system doesn't understand the text form it will use an image, while on systems that understand it it uses the Unicode code point.

Arcanemagus: You should _always_ use the text form wherever possible in Markdown documents as that allows _any_ text editor to work on the document, and systems without knowledge of the specific code point can still understand the document.

CONTRIBUTING.md
skylize: Initially I was trying to keep it pretty short to match the brevity of all the other steps. But after further editing, some of the other steps got longer too, which means some more text would not be out of place anymore. So how about this?

> Run `apm link --dev` to link your working folder as a development-mode-only package. Atom will load the package as if you installed it to `{user_home_directory}.atom\dev\packages\`.

Arcanemagus: The second part should be something like "Atom will load this development version instead of any regularly installed package, allowing you to test your code in a live environment."

(You can't install a package in `~/.atom/dev/packages`, that's _only_ for testing development packages, including overriding any version that is currently installed in `~/.atom/packages`.

CONTRIBUTING.md
skylize: So what instruction would you suggest on this? I mean you can rebase directly off `upstream master` to a branch the same as rebasing off a local `master`. If there is a merge conflict, it's gonna bite you either way. And merging instead of rebasing adds useless merge commits.

CONTRIBUTING.md
skylize: Works for me.

CONTRIBUTING.md
Arcanemagus: Yea, for somebody familiar with `git` it's not really that different. For everyone that I've seen submitting a PR off of `master`? It's incredibly difficult. Then again they are unlikely to read the instructions here so...

Anyway, I guess I'd write something like this?
```markdown
While on the `master` branch run `git checkout -b my-super-special-branch` to create and checkout a separate branch for your work. This makes it easier to rebase your work later if `master` updates before it gets merged.
```

CONTRIBUTING.md
```diff
@@ -4,47 +4,74 @@
...
+* Vertical whitespace helps readability, don’t be afraid to use it!
 
 ## Testing
 
-You can run the specs for this package locally by either running `apm test` from
-the project directory, or from within Atom by going to View -> Developer -> Run
-Package Specs.
+* Running `apm link --dev` will link your working folder to Atom as a development-mode-only package. Atom will load this development version of the package instead of any regularly installed package, allowing you to test your code in a live environment. Use `atom --dev some/project/folder` to start in developer mode.
+
+* There are several ways to start the specs. Any of these will run the full test suite inside the Atom environment. Make sure the tests pass before submitting a pull request.
```
Arcanemagus: You know what, how about we just link to [here](http://flight-manual.atom.io/hacking-atom/sections/package-word-count/#running-tests) in the Flight Manual, and let Atom handle documenting how to run the specs?

skylize: I think  linking to flight manual iss a nice idea, but I'm looking at what you pointed to and it's not very informative. I could see linking to the flight manual in general, but not for that.

PR #1092, opened 2018-01-24
Add ramda, fast-deep-equal, a few generic utils

skylize:


spec/f-utils/deep-diff-dict-spec.js
```diff
@@ -0,0 +1,52 @@
+'use babel'
+
+// eslint-disable-next-line no-unused-vars
+import 'jasmine-fix'
```
Arcanemagus: 🔥: not needed here.

spec/f-utils/f-utils-spec.js
```diff
@@ -0,0 +1,140 @@
+'use babel'
+
+// eslint-disable-next-line no-unused-vars
+import 'jasmine-fix'
```
Arcanemagus: 🔥: Not needed here.

spec/f-utils/f-utils-spec.js
```diff
@@ -0,0 +1,140 @@
...
+  isNotNil,
+  deepProp,
+  keyedFilter,
+  keyedMap,
+} from '../../src/f-utils'
+
+
+describe('fromMapToObj', () => {
+  it('transforms a Map to an Object dictionary', () => {
+    const map = new Map([['a', 'some string'], ['b', 10], ['c', {}]])
```
Arcanemagus: Values: `null`, `undefined`, `[]`?

Keys: All the things that Map's can use as keys but Arrays can't. (Mostly kidding, but what _should_ it do here?)

src/f-utils/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+
+/** ************
+ * Predicates *
+ ************* */
+
+// False if parameter is null or undefined, else true
+//
+// isNotNil :: Any -> Boolean
+export const isNotNil = r.pipe(r.isNil, r.not)
```
Arcanemagus: I was going to say this JavaScript not LUA... but that's the term Ramda uses for some bizarre reason 😮.

spec/f-utils/f-utils-spec.js
skylize: Hmm. Didn't consider that because I can't think of any reason I would want to use `null`, `undefined` or `[]` as a key.

skylize: What do _you_ think it should do here?

src/f-utils/index.js
skylize: I think it's nice to have a term that refers to both undefined and null, since they are so often used interchangeably.

spec/f-utils/f-utils-spec.js
Arcanemagus: Oh, I meant those as values, and if you have no use for them (and can't use them anyway) I would `throw`.

PR #1095, opened 2018-01-25
Some miscellaneous steps toward modularizing

skylize:
This is *very much* a work in progress. Short term goal is to break the existing codebase down into "modules" of related behaviors. Groupings are not set in stone, just choices that appear reasonable while trying to unwind the codebase. Some minor tweaks along the way, such as favoring function expressions over declarations.

### DIFF
--- a/src/config-inspector/index.js
+++ b/src/config-inspector/index.js
@@ -1,0 +1,4 @@
+'use babel'
+
+// eslint-disable-next-line import/prefer-default-export
+export { default as isLintDisabled } from './is-lint-disabled'
--- a/src/config-inspector/is-lint-disabled.js
+++ b/src/config-inspector/is-lint-disabled.js
@@ -3,7 +3,6 @@
 import { getConfigPath } from '../file-system'
 import isAtHomeRoot from './is-at-home-root'
 
-// eslint-disable-next-line import/prefer-default-export
 const isLintDisabled = ({ fileDir, disableWhenNoEslintConfig }) => {
   const configPath = getConfigPath(fileDir)
   const noProjectConfig = (
--- a/src/config-inspector/is-lint-disabled.js
+++ b/src/config-inspector/is-lint-disabled.js
@@ -1,0 +1,16 @@
+'use babel'
+
+import { getConfigPath } from '../file-system'
+import isAtHomeRoot from './is-at-home-root'
+
+// eslint-disable-next-line import/prefer-default-export
+const isLintDisabled = ({ fileDir, disableWhenNoEslintConfig }) => {
+  const configPath = getConfigPath(fileDir)
+  const noProjectConfig = (
+    configPath === null
+    || isAtHomeRoot(configPath)
+  )
+  return noProjectConfig && disableWhenNoEslintConfig
+}
+
+export default isLintDisabled
--- a/src/file-system/index.js
+++ b/src/file-system/index.js
@@ -1,0 +1,4 @@
+'use babel'
+
+// eslint-disable-next-line import/prefer-default-export
+export { default as getConfigPath } from './get-config-path'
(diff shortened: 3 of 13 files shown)


src/worker/task.js
```diff
@@ -0,0 +1,55 @@
...
+
+const createWorkerTask = () => {
+  let task = null
+  let started = false
+  const isConnected = () => !!(task && task.childProcess.connected)
+
+  /**
+   * Start the worker process if it hasn't already been started
+   */
+  const start = () => {
```
Arcanemagus: These functions are being redeclared each time the function is called, why not move them out to the file scope along with the  associated variables?

src/file-system/index.js
```diff
@@ -0,0 +1,4 @@
+'use babel'
+
+// eslint-disable-next-line import/prefer-default-export
```
Arcanemagus: Again, personally I'd split this up so you don't have to disable the rule and future work is simpler.

skylize: And again this file will be exclusively (or close to it) exporting other files to create a single entry point to an api of related code, so the future work is just deleting the `eslint-disable`.

src/config-inspector/is-lint-disabled.js
```diff
@@ -0,0 +1,16 @@
+'use babel'
+
+import { getConfigPath } from '../file-system'
+import isAtHomeRoot from './is-at-home-root'
+
+// eslint-disable-next-line import/prefer-default-export
```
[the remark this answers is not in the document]
skylize: Which are you asking about? For the name `isLintDisabled`, you will see the export is on `default` at the bottom. For the name, `getConfigPath` you will see it's imported from `file-system` which planned to hold more names. The `eslint-disable` is a stray that's not doing anything so we can burn that.

src/config-inspector/index.js
```diff
@@ -0,0 +1,4 @@
+'use babel'
+
+// eslint-disable-next-line import/prefer-default-export
```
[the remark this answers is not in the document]
skylize: This file is intented to be a flat api for the code in this folder. It's a single entry point for consumers. This file will mostly (or only) be exporting other files and not doing any work of its own.

src/worker/task.js
skylize: This is creating what is, for all intents, an `instance`. Call `createWorkerTask` and it wraps all these variables in a closure creating state for a worker and returning an object with the api to interact with it. Then if you choose to call `createWorkerTask` again, you get a new closure, with new variables, and a new worker task. If you use only 1 worker, then you only call it once, and it doesn't matter that it's recreating. If we want to grab a backup worker in case of failure, then this (isn't quite there yet but) is moving in the direction of making that easy.

src/config-inspector/is-lint-disabled.js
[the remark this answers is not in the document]
Arcanemagus: Ah, I was wondering why you were disabling it here, but not in `file-system/get-config-path`. 👍

src/worker/task.js
Arcanemagus: I know you hate them, but that's _exactly_ what a `class` is for (and _also_ stops re-declaring the methods on each call). 😛

skylize: Thanks for the input. I'll leave it like this for now and take that into consideration as I move forward. Mostly it was just a way to isolate these worker calls away from the stuff using it so I could get it into a separate file. It gets that job done, so I can keep looking for other pieces to break out.

At the current stage, I'm trying to minimize changes to existing code beyond breaking it into smaller pieces.

The rest of the code currently accesses the worker through method calls on the task, so this allowed them to keep doing that, but with one layer of abstraction in between.  In the future, I intend to look at making this more functional and declarative, which might, for example, look something like this.

```
const jobPath = './job'

const getRunningTask = src => task =>
   deepProp('childProcess.connected', task)
    ? task
    : new Task(require.resolve(src))

const getRunningJobTask = getRunningTask(jobPath)

const observeTask(task, event, f) => {
  task.on(event, f) // ewe, side effect :(
  return task
}
```

PR #1096, opened 2018-01-26
Add debug & linter-report modules. No more helpers.

skylize:
This patch completely eliminates what was once `helpers.js`, having now spread all of the code from that file across numerous other modularized files.

The short term goal is still to break apart the codebase into small submodules of related behavior. Taking care at this stage to resist making significant changes to the existing code, unless required to get decoupling. Some minor changes applied along the way, mostly stylistic, such as moving to consistently favor function expressions over declarations, and lots of renaming of functions to fit module groupings.


skylize:
Sorry, wrong branch.

PR #1097, opened 2018-01-26
Add debug & linter-report modules. No more helpers.

skylize:
This patch completely eliminates what was once helpers.js, having now spread all of the code from that file across numerous other modularized files.

The short term goal is still to break apart the codebase into small submodules of related behavior. Taking care at this stage to resist making significant changes to the existing code, unless required to get decoupling. Some minor changes applied along the way, mostly stylistic, such as moving to consistently favor function expressions over declarations, and lots of renaming of functions to fit module groupings.


src/debug/construct.js
```diff
@@ -0,0 +1,20 @@
+'use babel'
+
+import debugInfo from './info'
+
+const constructDebug = async () => {
```
Arcanemagus: Debug _what_? `Object`? `Array`? (There's a reason the original function was `generateDebugString`)

src/linter-report/simple.js
```diff
@@ -0,0 +1,28 @@
...
+const simpleReport = (textEditor, {
+  severity = 'error',
+  excerpt = '',
+  description,
+} = {}) => [{
+  severity,
+  excerpt,
+  description,
+  location: {
+    file: textEditor.getPath(),
```
Arcanemagus: 📝 Why aren't we just using the path from ESLint? Just a note here, I realize the old code was already doing it this way.

Arcanemagus:
AppVeyor is mistakenly associating this branch with #1096, attempting a close/reopen to bring it to its senses.

Arcanemagus:
Just ignore AppVeyor...

src/linter-report/simple.js
skylize: I don't know. Sounds like a good suggestion to me.

src/debug/construct.js
skylize: I don't typically expect the datatype of a return value to be part of a function name, unless it distinguishes it from a similar function with a different return type.

Is there alternatively a debug object or a debug array that I missed? When we want "A Debug" to provide to the user, we call this function. The `debugInfo` submodule returns and object, and you aren't requesting to rename that to `debugInfoObject`.

I think a better solution here is to get this `flow` thing in place and define a return type.

skylize:
I think "messages" is an overly common term and thus excessively vague. I chose `report` because I was looking for something succinct that could clearly distinguish this final processed output passed out to `linter` on each lint from any other "message" or "lint" or anything like that which might appear in the codebase. 

Just look at this definition of the return value from `processJobResponse` to easily see how our overuse of the word message has already become somewhat convoluted and confusing.
```
* @return {Promise}               The messages transformed into Linter messages
```
Ummm. What? Let's try that again.
```
* @return {Promise}               Eslint messages transformed into an atom-linter report.
```
Ahh, much better. Seems clear to me that finding alternative names for at least some types of "messages" or message containing objects is a huge win. I am open to other suggestions. I will certainly make a point of ensuring any naming inconsistencies caused by this get corrected.

src/debug/construct.js
skylize: I think having `string` in the name was more important when this was all flat files. Much easier to get it mixed up with the `getDebugInfo`, which was in the same file and in the same scope and returns an object. Now, `debug.construct` has a clearly hierarchically superior position over its submodule.

skylize: Hmm. Leaving the old arguments up partly for posterity and partly because they still make some useful points. But I'm looking closer at the code and starting to rethink this a bit.

The object returned from `debug.info` really is everything that makes up a debug, except for some labeling for string output. Perhaps, `debug.userMessage`? or something like that. I guess the target is us, more-so than the user, so maybe not quite right. But namely I'm _much_ more interested in a name that captures its intent of prepping it for display, than describing it as a `string`.

Arcanemagus: `debugMessage`?

I like the old name because it says _exactly_ what it is doing, when I saw `constructDebug` I had no idea what it was doing until I looked at the code and saw what it used to be.

`debug.label` doesn't work since this isn't just a label but the entire message.

skylize:
Oh wow. That does create some pretty strong cognitive disonance. 🤔 

We have too many things all vying for the same namespace. `lint`, `linter`, `message`, `messages`. How do we get out of this?

Arcanemagus:
> We have too many things all vying for the same namespace. `lint`, `linter`, `message`, `messages`. How do we get out of this?

If you need to clarify, just add the name of what the message belongs to?
`eslintMessages`, `linterMessages`, etc.

skylize:
But ESLint is a Linter. So the name `linterMessage` to say it is _not_ a message from ESLint is still wildly confusing.

Arcanemagus:
Specificity rules here, although ESLint belongs to the _class_ of programs called "linter", those messages are meant for a program that is _literally_ called `linter`.

skylize:
Yeah. As demonstrated by the final commit on this branch, I'm definitely conceding on this issue (at least until/unless I can come up with something better that still satisfies what you want from of the name).

PR #1101, opened 2018-02-09
Break apart worker-helpers

skylize:
Worker-helpers file eliminated, with all functions spread out to other files. Split worker file to 2 directories between management and child process. Implement thin-layer subscription api in between send-job and actual Task subscriptions, reducing Task listeners to only 2 long-running subs (success/fail) per worker. Miscellaneous other tweaks, mostly focused around the Worker.


skylize:
Sorry. I did it again. Not intended for master branch. 🙈

PR #1103, opened 2018-02-10
Enforce Node 7.4.0, remove transpile (except for specs)

skylize:
Add eslint-node plugin to enforce no-unsupported features at version 7.4.0. Remove all use of import/export and async/await to comply. Disable rule for specs, allowing tests to continue using async/await. Remove 'use babel' transpile flag from all src files.


skylize:
Further discussion indicates this is not a sufficiently useful change to transpiling behavior. Closing as invalid.

PR #1104, opened 2018-02-11
Respectful idleCallback loader and small CompositeDisposable

skylize:
Add an idleCallback loader that respects deadline from scheduler, requesting new idleCallback when out of time. Use idleCallbacks to lazy require packages only for caching purposes, rather than saving to variables. Use disposables interface for disposing of idleCallbacks. Add small makeCompositeDisposable function to avoid requiring from 'atom' to get the built-in class.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,61 +1,36 @@
 'use babel'
 
-// Dependencies
 // NOTE: Requiring files adds significant time to package startup. To avoid this
 // delay, we are *not* requiring at the top of this file. Use idleCallbacks to
-// attempt to pre-cache requires for fast resolution if scheduler allows. Since
-// all requires are statically defined and inside some function, Atom should be
-// able "snapshot" the package.
-//
-const dependencies = [
-  () => require('./validate/editor'),
-  () => require('./linter-message'),
-  () => require('./worker-manager'),
-  () => require('./eslint-config-inspector'),
-  () => require('./debug'),
-  () => require('./rules'),
-  () => require('path'),
-  () => require('./atom-config'),
-]
-
-// Miscellaneous idle tasks.
-//
-const idleTasks = [
-  // Pre-start the worker
-  () => require('./worker-manager').task.start(),
-  // Install peer packages
-  () => require('atom-package-deps').install('linter-eslint'),
-]
+// attempt to pre-cache requires for fast resolution if scheduler allows.
 
 // Run idle tasks with all reasonable attempts to avoid blocking.
 //
-const makeIdleRunner = () => {
+const linterEslintMakeIdleRunner = () => {
   const { requestIdleCallback, cancelIdleCallback } = window
   let id
-  let tasks
+  let tasks = []
 
-  const loadLazily = (deadline) => {
+  const linterEslintScheduleIdle = (deadline) => {
     // While time alotted by scheduler and work remaining
     while (deadline.timeRemaining() && tasks.length) {
       // Remove and run function from end of list.
       tasks.pop()()
     }
-
     // If work remains, request more time
-    if (tasks.length) id = requestIdleCallback(loadLazily)
+    if (tasks.length) {
+      id = requestIdleCallback(linterEslintScheduleIdle)
+    }
   }
 
-  return (requestedTasks) => {
-    // Spread avoids external mutation and allows non-array iterable
-    tasks = [...requestedTasks]
-    // Init first request
-    id = requestIdleCallback(loadLazily)
-    // Return a disposable
+  return (requestedTasks = []) => {
+    tasks = [...requestedTasks, ...tasks]
+    id = requestIdleCallback(linterEslintScheduleIdle)
     return { dispose: () => cancelIdleCallback(id) }
   }
 }
 
-// // Simplified composite disposable avoids requiring atom core in main
+// // Simplified composite disposable
 // //
 const makeCompositeDisposable = () => {
   const disposables = new Set()
@@ -73,18 +48,36 @@ const makeCompositeDisposable = () => {
 
 module.exports = {
   activate() {
-    // const { CompositeDisposable } = require('atom')
-    // this.subscriptions = new CompositeDisposable()
-    this.subscriptions = makeCompositeDisposable()
-
-    // Put dependency loading on Atom's todo-list
-    this.subscriptions.add(makeIdleRunner()(dependencies))
-
     const {
       atomConfig,
       getMigrations,
       subscribe: configSubscribe
     } = require('./atom-config')
+
+    const unlessSpec = f => () => !atom.inSpecMode() && f()
+
+    const idleTasks = [
+      // Migrate any outdated config settings
+      () => getMigrations(),
+      // Start the worker Task process
+      unlessSpec(() => require('./worker-manager').task.start()),
+      // Install peer packages
+      unlessSpec(() => require('atom-package-deps').install('linter-eslint')),
+      // Pre-cache dependency requires
+      () => require('./validate/editor'),
+      () => require('./linter-message'),
+      () => require('./worker-manager'),
+      () => require('./eslint-config-inspector'),
+      () => require('./debug'),
+      () => require('./rules'),
+      () => require('path'),
+    ]
+
+    this.subscriptions = makeCompositeDisposable()
+
+    // Put idle tasks on Atom's todo-list
+    const requestIdles = linterEslintMakeIdleRunner()
+    this.subscriptions.add(requestIdles(idleTasks))
 
     // Subscribe to Atom configuration settings
     this.subscriptions.add(...(configSubscribe()))
@@ -129,8 +122,8 @@ module.exports = {
           if (!activeEditor) return false
 
           // Some scary voodoo black magic! Atom v1.19.0+
-          // Compares the private component property of the active TextEditor
-          //   against the components property of the TextEditor DOM elements
+          // Compare private component property of active TextEditor to the
+          // component props of elements to find matching <TextEditor> DOM Node.
           const evtIsActiveEditor = evt.path.some(elem =>
             (elem.component && activeEditor.component
               && elem.component === activeEditor.component))
@@ -143,12 +136,6 @@ module.exports = {
         }
       }]
     }))
-
-    // Load miscellaneous idle tasks and potential settings migrations
-    if (!atom.inSpecMode()) {
-      this.subscriptions.add(makeIdleRunner()(getMigrations()))
-      this.subscriptions.add(makeIdleRunner()(idleTasks))
-    }
   },
 
   deactivate() {
--- a/src/main.js
+++ b/src/main.js
@@ -3,9 +3,9 @@
 // Dependencies
 // NOTE: Requiring files adds significant time to package startup. To avoid this
 // delay, we are *not* requiring at the top of this file. Use idleCallbacks to
-// attempt to pre-cache requires for fast resolution if scheduler allows.
-// Since all requires are statically defined and inside some function,
-// Atom should be able snapshot our package.
+// attempt to pre-cache requires for fast resolution if scheduler allows. Since
+// all requires are statically defined and inside some function, Atom should be
+// able "snapshot" the package.
 //
 const dependencies = [
   () => require('./validate/editor'),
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -1,144 +1,175 @@
...
+  () => require('./eslint-config-inspector'),
+  () => require('./debug'),
+  () => require('./rules'),
+  () => require('path'),
+  () => require('./atom-config'),
+]
+
+// Miscellaneous idle tasks.
+//
+const idleTasks = [
```
Arcanemagus: This is a static list that is only ever used in one place, just embed it there.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
-    knownRules = require('./rules').default
-  }
-}
+// NOTE: Requiring files adds significant time to package startup. To avoid this
+// delay, we are *not* requiring at the top of this file. Use idleCallbacks to
+// attempt to pre-cache requires for fast resolution if scheduler allows. Since
+// all requires are statically defined and inside some function, Atom should be
+// able "snapshot" the package.
+//
+const dependencies = [
```
Arcanemagus: Static list, just declare it where it is getting used.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
+]
+
+// Run idle tasks with all reasonable attempts to avoid blocking.
+//
+const makeIdleRunner = () => {
+  const { requestIdleCallback, cancelIdleCallback } = window
+  let id
+  let tasks
+
+  const loadLazily = (deadline) => {
```
Arcanemagus: This isn't _just_ being used to lazily load dependencies. Since this is the name that will show in CPU profiles and we want some way of identifying it how about `linterEslintWork`?

src/main.js
```diff
@@ -1,144 +1,175 @@
...
   }
+}
 
-  if (!atom.inSpecMode()) {
-    makeIdleCallback(linterEslintInstallPeerPackages)
-    makeIdleCallback(linterEslintLoadDependencies)
-    makeIdleCallback(linterEslintStartWorker)
+// // Simplified composite disposable avoids requiring atom core in main
+// //
+const makeCompositeDisposable = () => {
```
Arcanemagus: What is your reasoning for needing to write this? In chat you briefly mentioned it was "faster", but I'd have to see some pretty big performance gains from this to justify writing our own version instead of just using what Atom already has written and maintains for us.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
+    dispose: () => disposables.forEach(d => d.dispose())
   }
 }
 
+
 module.exports = {
   activate() {
-    getMigrations().map(makeIdleCallback)
-    this.subscriptions = new CompositeDisposable()
+    // const { CompositeDisposable } = require('atom')
```
Arcanemagus: 🔥 If we _are_ switching, we don't need commented out code here...

src/main.js
```diff
@@ -1,144 +1,175 @@
...
       'atom-text-editor:not(.mini), .overlayer': [{
         label: 'ESLint Fix',
         command: 'linter-eslint:fix-file',
         shouldDisplay: (evt) => {
           const activeEditor = atom.workspace.getActiveTextEditor()
-          if (!activeEditor) {
-            return false
-          }
-          // Black magic!
+          if (!activeEditor) return false
```
Arcanemagus: I personally absolutely hate inline `if` statements, I thought the ESLint config actually blocked them...

Arcanemagus: (At least JS doesn't allow _postfix_ control structures like some other languages...)

src/main.js
```diff
@@ -1,144 +1,175 @@
...
+
+          // Some scary voodoo black magic! Atom v1.19.0+
           // Compares the private component property of the active TextEditor
-          //   against the components of the elements
+          //   against the components property of the TextEditor DOM elements
```
Arcanemagus: This actually makes the comment wrong, there are _many_ DOM elements that are being compared against, only one of which is likely to be a `<TextEditor>`.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
           const { scopes } = atomConfig
+          const { hasValidScope } = require('./validate/editor')
+
           return evtIsActiveEditor && hasValidScope(activeEditor, scopes)
         }
       }]
     }))
 
-    scheduleIdleTasks()
+    // Load miscellaneous idle tasks and potential settings migrations
```
Arcanemagus: This is actually _performing_ the migrations, not loading them.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
 
   deactivate() {
-    idleCallbacks.forEach(callbackID => window.cancelIdleCallback(callbackID))
-    idleCallbacks.clear()
-    if (worker) {
-      // If the helpers module hasn't been loaded then there was no chance a
-      // worker was started anyway.
-      worker.task.kill(true)
-    }
+    const { task } = require('./worker-manager')
```
Arcanemagus: This is an example of why I don't like (ab)using `require`, this will do work (all the setup you have being done when `worker-manager` is `require`d) just to see if work was already done and it needs to get rid of it.

src/main.js
```diff
@@ -1,144 +1,175 @@
...
+    if (task) task.kill(true)
     this.subscriptions.dispose()
   },
 
   provideLinter() {
+    const {
```
Arcanemagus: If you are going to abuse `require`, _own_ it and just `require` this stuff before it is used.

src/main.js
skylize: That's a good point. When I came up with that name, it was only being used for deps, so the name was pretty clear. But I do think something on the lines of "lazy" or "idle" is important to expressing intention.... Um. Maybe just `runLazily`? Not sure.

src/main.js
skylize: Sorry, I don't understand. Could you rephrase?

skylize: Oh, I think I get it after seeing you next same request. This is similar to listing imports at the top of the file. Even though we are not actually _doing_ the import, it still gives an overview of what's coming in. Additionally it separates data from behavior. It' much easier to comprehend the data (a list of dependencies) when not mixed in with behavior (lazily attempting to import files), and it is equally easier to comprehend the behavior when not having to parse through a list of files.

src/main.js
skylize: I came up with it when converting to all `require`s instead of `import`s. 

Of course that plan has been scrapped, but when using require, it's easier to notice that in order to import `CompositeDisposable` the entire `atom` entry point with tons of classes we have no use for must be parsed through to dig out the class we want.

I had just recently before that worked out what that class was actually doing (basically just holding a list of things with a `dispose` method and calling said method for everyone in the list). So I threw together a function that provided the same basic behavior just to see.

I don't know what you're perfs will show when you manage to zoom in on our startup times. But I put a lot of tries into testing this across varying loads, with only Atom running or with tons of apps open consuming my CPU. 

Except for some very occasional and totally random outliers that went the other direction, this change under all cases (at least on my system) results in a minimum 6% speed improvement for running the full test suite, usually closer to 10%. 

So for example, if my CPU load was finishing the suite in 80s, I could make the tiny change of swapping this in for the class import and finish the test suite in 72s under the same load, pretty consistently.

src/main.js
skylize: Yep. artifact from my A/B testing.

src/main.js
skylize: No. The config allows 1-liners, but forces braces if you do 2 lines.

I'm not really a huge fan of 1-liners either, very much preferring (in most cases) 2 lines with just an indent and no braces. But I quite strongly dislike using _3_ lines for a statement that is very readable in only 1. And I find braces to be just meaningless noise, much like semicolons. So that was my compromise to comply with the linting rules while reducing useless noise.

Didn't know you had a hatred for them. I guess I'll just quit doing that? 😕

src/main.js
skylize: Ok. How can we make this accurate without just reverting? I think not mentioning the DOM is a pretty big commission in this comment.

src/main.js
Arcanemagus: > Even though we are not actually doing the import, it still gives an overview of what's coming in.

You're abusing `require` all through the rest of the file, so this no longer has any meaning.

----

I'm just saying move this (and the next) block down to right before it is actually used. That way it is scoped into the `activate` and gotten rid of when no longer used, as well as having the advantage that you don't have to go searching for what  those lines are actually doing when you see them.

src/main.js
Arcanemagus: The fact that it is running in an idle callback is enough for my concerns, no matter what `linterEslint` needs to be in the name of something _pretty_ high up the stack trace so you can trace back where this is coming from.

src/main.js
skylize: Well this line of code is definitely _not_ performing any migrations. It's simply passing a list of to the idleCallback runner function. I think in my head it was "loading the list of migrations into the runner". Happy to hear how you might word it.

src/main.js
Arcanemagus: > it's easier to notice that in order to import CompositeDisposable the entire atom entry point with tons of classes we have no use for must be parsed through to dig out the class we want.

`atom` is embedded in the VM, the only cost to `require`ing it is resolving the package name. You can see this yourself in the profile. If there is any difference in the performance it's in the implementation of the `CompositeDisposable` itself, which would make me question what is so different that it is that much faster.

src/main.js
skylize: Yeah, I spotted that, and had a little icky feeling about it for a very brief moment. But let's be real here. Except when running specs, how often are we going to reach `deactivate` before we've had a chance to load the task? Basically never.

src/main.js
Arcanemagus: "elements" -> "DOM elements"?

src/main.js
Arcanemagus: There's a reason I sent this as a comment, not requesting changes. This is a style change that can get sorted out in the eventual final merge, but I figured I would mention it since you have started using it more and more.

src/main.js
skylize: That's about readability. Truly-inline requires can get pretty noisy. Once we reach the top of that function we know for certain we need it now. The one you chose to comment on is particularly notable for the fact that we also need to dig in for more nested data. 

```
showRule: require('./atom-config').atomConfig.showRule
```

The main point of the "inline" requiring is so that we can get all the imports inside of functions, thus avoiding Atom trying to parse the import files right away, and also making it possible for Atom to make snapshots for even faster loads. Once we are inside a function, it mostly doesn't matter if we import at the top of the function or later on (except for in the very first `activate` function, where we are just hopeful we manage to _somehow_ get at least 1 or 2 of them loaded before we get there. 

The point of using `require` instead of variables is 1.because its a lot easier to see what's actually going on when you don't have this out-of-nowhere function that magically makes unspecified names suddenly available that weren't usable before calling it.

And 2. It's easier to idly load, when you don't have to worry about whether the idle loading was in-time or not. We just say "load this when you can", and if it doesn't happen, so what. "We'll just grab this one ourselves, and you load the next one."

This method tries to treat it as close as possible to the more common `import`s. We're just importing at the top of the function instead of at the top of the file. 

The `require` function has a certain behavior, which includes caching. I don't think taking advantage of that behavior when we're pretty much _forced_ to use it in place of `import` qualifies as "abusing" it.

src/main.js
skylize: Please go ahead and pull this to your machine and do some of your own A/B testing and check it out in perfs. If there is no cost to the import then I really don't know where the performance is coming from. 

Maybe it is a fluke on my machine. Maybe `CompositeDisposable` is overly complicated for its job, or poorly written, or written to handle a bunch of edge cases we don't care about. All I know at this point is I saw significant and consistent speedup of the test suite that directly correlates with this change.

src/main.js
skylize: No. It was worth mentioning. If it's gonna get burned anyway, I might as well just get used to it.

src/main.js
Arcanemagus: > making it possible for Atom to make snapshots for even faster loads.

This is an _incredibly_ long term goal btw, snapshots are currently only done for packages bundled into Atom itself.

> except for in the very first activate function, where we are just hopeful we manage to somehow get at least 1 or 2 of them loaded before we get there.

😮 Uh.... that doesn't sound good _at all_, previously dependencies were only needed in commands that could only be triggered by the user, and even then they were ensured to be loaded by the time they were used. We should _never_ be "hopeful we manage to somehow get at least 1 or 2 of them loaded before we get there".

src/main.js
skylize: > "elements" -> "DOM elements"?
This makes sense, except unfortunately "elements" is used much more widely than just the DOM. I do not think we need to mention the dom in the code, but if we are going to take the time to write comments about it, that's a pretty bit point bring up. It's basically the only place in the code we touch the dom, so it's pretty significant that we are suddenly interacting with it. (It's only 3 letters long. 😄 ).

src/main.js
skylize: Only hopeful in that we hope the scheduler beats us to it while we are in the middle of something else anyway. Either way, when we actually get to where the code is needed, we will get that code. There is nothing questionable about that.

src/main.js
skylize: Not a fan of what that does to the visual organization of the code, but you make a fair point. My plan is to concede this one without further argument.

PR #1106, opened 2018-02-13
Add hex-helix logo to markdown headers

skylize:
Add logo svg file, and use that image in H1 headers of markdown files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Changelog
+# <img alt="linter-eslint logo" src="logo.svg" width="48px" height="48px"/> Changelog
 
 ## v8.4.1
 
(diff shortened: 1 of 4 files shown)


Arcanemagus:
So this is _awesome_, but... I have _no_ idea if we can use it since it might fall under a "modified" Atom logo provision, and that follows the terms of https://github.com/logos.

skylize:
The artwork is totally original. Strongly inspired by, but _clearly_ not a copy of the Atom logo. I think it would hold up in court as an acceptable derivative work. Atom is certainly not the first use of a helix. And this logo actually looks more like a flower than a helix. And on top of that, the package is completely devoted to making Atom better at no cost.

Is there someone we could contact to get clearance? That seems a lot easier than wondering anxiously if it is a problem or hiring lawyers.

PR #1112, opened 2018-02-17
Remove bundled fallback ESLint

skylize:
Remove fallback to the bundled copy of ESLint. Make findEslintDir and debug.info more declarative. Move some ESLint related utils to worker-process folder.


aminya (changes requested):
Bundling Eslint ensures that the package is functional across all the systems no matter if a local or global Eslint is provided and without the need for doing additional work. Obviously, the users can configure things to use a local/global installation, but we don't need to force them to do so.

PR #1116, opened 2018-03-02
Fix CI builds

Arcanemagus:
Update the version of `eslint-plugin-import` used internally in the specs to the one that is installed by `apm` (which ignores `package-lock.json` currently), also updates the specs to accommodate the changes in the latest version.

### DIFF
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -192,12 +192,12 @@ describe('The eslint provider for Linter', () => {
       const editor = await atom.workspace.open(paths.badImport)
       const messages = await lint(editor)
       const expected = "Unable to resolve path to module '../nonexistent'. (import/no-unresolved)"
-      const expectedUrl = 'https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md'
+      const expectedUrlRegEx = /https[\S]+eslint-plugin-import[\S]+no-unresolved.md/
 
       expect(messages.length).toBe(1)
       expect(messages[0].severity).toBe('error')
       expect(messages[0].excerpt).toBe(expected)
-      expect(messages[0].url).toBe(expectedUrl)
+      expect(messages[0].url).toMatch(expectedUrlRegEx)
       expect(messages[0].location.file).toBe(paths.badImport)
       expect(messages[0].location.position).toEqual([[0, 24], [0, 40]])
       expect(messages[0].solutions).not.toBeDefined()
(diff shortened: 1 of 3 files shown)


Arcanemagus:
Travis CI failed that build so badly I can't even restart it...

spec/linter-eslint-spec.js
```diff
@@ -192,12 +192,12 @@ describe('The eslint provider for Linter', () => {
       const editor = await atom.workspace.open(paths.badImport)
       const messages = await lint(editor)
       const expected = "Unable to resolve path to module '../nonexistent'. (import/no-unresolved)"
-      const expectedUrl = 'https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md'
+      const expectedUrlRegEx = /https:\/\/.+\/eslint-plugin-import\/.+\/no-unresolved.md/
```
skylize: I'm not a fan of using regexp `.+` or `.*` when it's not really necessary, since it can literally match anything. This string could have spaces and new-lines and a bunch of weird symbols in the middle of the "url", and it would all be fine. How about `[/\w-.]+` instead of `.+`? This matches on 1 or more of forward-slash, any alpha-numeric, underscore, dash, and dot.

With that change in place, I think it is reasonably sane if you want to simplify by removing explicit checks for `/`, in which case the full expression would read 
```javascript
/https:[/\w-.]+eslint-plugin-import[/\w-.]+no-unresolved.md/
```
which in my opinion is actually even slightly easier to read than the original pattern, despite the more complex character class description.

Otherwise, everything looks good to me.

spec/linter-eslint-spec.js
```diff
@@ -655,7 +656,7 @@ describe('The eslint provider for Linter', () => {
   })
 
   describe('handles the Show Rule ID in Messages option', () => {
-    const expectedUrl = 'https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md'
+    const expectedUrlRegEx = /https:\/\/.+\/eslint-plugin-import\/.+\/no-unresolved.md/
```
skylize: Same as comment above about `.+` regexp.

spec/linter-eslint-spec.js
Arcanemagus: > This string could have spaces and new-lines and a bunch of weird symbols in the middle of the "url", and it would all be fine.

We aren't testing the validity of the URL they are using, that's their job. We just care that it "looks like" a URL for that rule.

That being said, there isn't a reason _not_ to change that, although I'm going to use `\S` instead of `\w` since URLs are _far_ less restrictive than `\w`.

skylize: Still a little lose for my taste, but I think `\S` a fair compromise here. I definitely agree we're not really trying to prove its a URL. It's mostly a sanity check on the characters between what we're interested in.

Issue #845, opened 2017-03-09
Fallback to globally installed eslint if not installed in project


gnestor:
## Issue Type
Feature Request

## Issue Description
I have a global eslint setup with `~/.eslintrc` and globally installed eslint, babel-eslint, etc. This is great because I can now lint all JS projects, not just those that have eslint installed and/or configured. However, if I am working on a project that does have its own eslint setup, then I start running into errors like "eslint-plugin-flowtype not found" because it's required by the project's eslint config but not installed globally. 

I propose that this package offer a global eslint installation fallback option so that project installations take precedence. 

Thoughts?

Arcanemagus:
An option like the [Use Global ESLint](https://github.com/AtomLinter/linter-eslint/blob/v8.1.3/package.json#L18) option that already exists? 😛

As a note, from what you are describing the project in question _does_ have an ESLint configuration, so it _should_ be installed in the project.

IanVS:
@Arcanemagus, actually we don't 'fall-back'.  The `Use Global ESLint` forces the global version to always be used, regardless if the project has its own locally installed ESLint.

We have talked about making this change in the past, and agreed it would be worthwhile.  But so far, nobody has done the work and submitted a PR.

IanVS:
@gnestor I assume you have `Use global` checked, right?  Can you share the output of the `Linter Eslint: Debug` command?

gnestor:
@IanVS Currently, when I am working in a project that DOES NOT have eslint set up, I enable `Use global`. When I switch to a project that DOES have a eslint set up, I need to disable `Use global`. I am proposing that `Use global` only uses the global eslint installation if there is no local installation.

My debug:

```
Atom version: 1.14.4
linter-eslint version: 8.1.3
ESLint version: 3.17.1
Hours since last Atom restart: 14.3
Platform: darwin
Using global ESLint from: /usr/local/lib/node_modules/eslint
Current file's scopes: [
  "source.js.jsx",
  "meta.class.body.js",
  "meta.function.method.js",
  "meta.tag.jsx",
  "JSXAttrs",
  "JSXNested",
  "meta.embedded.expression.js",
  "source.js.jsx",
  "meta.tag.jsx",
  "JSXAttrs",
  "comment.line.double-slash.js"
]
linter-eslint configuration: {
  "disableWhenNoEslintConfig": false,
  "disableWhenNoEslintrcFileInPath": true,
  "lintHtmlFiles": true,
[...]

PR #1144, opened 2018-06-20
updating global linter as fallback

bcnichols3:
reposting my comment from [issue #845](https://github.com/AtomLinter/linter-eslint/issues/845)

Definitely. My original thinking is to make it so useGlobalEslint remains as an option to supersede local eslint? Though that sounds like a terrible idea. I wonder if that check should be removed entirely, so as to not make it easy to jump over a local eslint — which is definitely a bad idea.

Currently I have:

if (localNodeModules && !useGlobalEslint) then locationType = 'advanced specified'; and eslintDir uses a ternary to determine relative or absolute path.
else if (!useGlobalEslint) then locationType = 'local project' and eslintDir is pulled from the modules.
if (!isDirectory(eslintDir)) then locationType = 'global' and eslintDir is resolved to global directory, with respect to OS.
if (isDirectory(eslintDir)) then return the object
else then `throw new Error('ESLint not found, please add a local eslint to this project or ensure the global Node path is set correctly.')
But I can change that to reflect the thinking above. I also altered worker helper tests accordingly, but two eslint provider tests are failing:

when a file is specified in an eslintIgnore key in package.json it will not give warnings when linting the file is throwing the wrong error
it errors when no config file is found is also throwing the wrong error

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
       "title": "Global ESLint",
       "properties": {
         "useGlobalEslint": {
-          "title": "Use global ESLint installation",
+          "title": "Always use global ESLint installation",
           "description": "Make sure you have it in your $PATH",
           "type": "boolean",
           "default": false,
--- a/spec/worker-helpers-spec.js
+++ b/spec/worker-helpers-spec.js
@@ -62,14 +62,25 @@ describe('Worker Helpers', () => {
       expect(foundEslint.type).toEqual('global')
     })
 
-    it('falls back to the packaged eslint when no local eslint is found', () => {
-      const modulesDir = 'not/a/real/path'
-      const config = createConfig({ global: { useGlobalEslint: false } })
-      const foundEslint = Helpers.findESLintDirectory(modulesDir, config)
-      const expectedBundledPath = Path.join(__dirname, '..', 'node_modules', 'eslint')
-      expect(foundEslint.path).toEqual(expectedBundledPath)
-      expect(foundEslint.type).toEqual('bundled fallback')
-    })
+	it('falls back to the global eslint when no local eslint is found', () => {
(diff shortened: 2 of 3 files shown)


IanVS:
>make it so useGlobalEslint remains as an option to supersede local eslint? 

It seems to me that using a global eslint installation in place of a locally-specified one (the current behavior) is a bad idea.  It can cause errors when the local config relies on plugins, which in order to use the global installation also must be installed globally.  But then what if you are working on two projects, each relying on different versions of a particular plugin?  Or what if your version of eslint allows particular behavior but the project's package.json specifies a version which will flag that behavior as an error?  Generally, it's sucky for all the reasons that global installations of eslint are discouraged in the first place.  

It seems like the only really useful time that folks have said they want to have a global installation is when they're working on small scripts that aren't really part of a "project".  Or perhaps when working on eslint itself, when it's useful to `npm link` to create a global reference which points back to the source you're working on.  But in both cases, I believe simply having an option to `Fall back to global installation` would be sufficient, because a standard local installation won't be found in either situation.

So, what do you think about changing the option entirely, from `Force global installation to be used` to `Allow global installation to be fallen back to if a suitable local installation is not found`?  We're going to have a breaking release soon anyway, so this is probably the time to do it if so desired.  I believe this was also the solution requested and agreed upon in the linked issue #845.

bcnichols3:
checkin in @IanVS ... I have the option being used as a fallback as requested. One test failure for some reason; can't quite reason out why! Maybe you know?

skylize:
```
The eslint provider for Linter
  when setting `disableWhenNoEslintConfig` is false
    it errors when no config file is found
```

This spec is probably expecting to use the fallback-installation of ESLint, which is no longer accessible. There are probably more tests like that, which will yell at you after you fix this one.

PR #1181, opened 2018-10-09
Revert "Add 'javascript' scope by default"

IanVS:
This reverts commit 1f2186ce29ddc1e503f13ad4cd7e3c9425100dd5.

Tree sitter now uses the regular TextMate based scopes, so this is not required and could be confusing.


Arcanemagus:
For reference, this was reverted in Atom by https://github.com/atom/atom/pull/17881.

PR #1185, opened 2018-10-10
Add commitlint

Arcanemagus:
Add [commitlint](http://marionebl.github.io/commitlint) as well as a [Conventional Commits](https://www.conventionalcommits.org/) based configuration to enforce commit messages are always in a consistent format.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,9 +18,6 @@ defaults: &defaults
         name: APM version
         command: ${APM_SCRIPT_PATH} --version
     - run:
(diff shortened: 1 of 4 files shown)


Arcanemagus:
I _think_ we will need to override Renovate's default config for the commit message format here. The ESLint format is not at all as well supported under default configurations for most tools.

Arcanemagus:
This has been updated to use the [Conventional Commits](https://www.conventionalcommits.org/) standard as the ESLint standard isn't supported in many of the required tools.

Issue #1122, opened 2018-03-24
linter-eslint no longer reporting anything


bennypowers:
## Issue Type
Bug

## Issue Description
I've noticed that for the last week or so, I no longer get eslint messages in diagnostics. `ts` errors from ide-typescript still appear, and there are plenty of linting errors in my file (see screenshot). Apologies for the vague description but I'm so far unable to uncover more information. Could be related to https://github.com/AtomLinter/linter-eslint/issues/1029 but I don't have any error messages in the js console.

![screen shot 2018-03-24 at 19 34 31](https://user-images.githubusercontent.com/1466420/37866960-6512d31a-2f9a-11e8-9e11-961f0c406509.png)

## Bug Checklist
- [x] Restart Atom
- [x] Verify the `eslint` CLI gives the proper result, while `linter-eslint` does not
- [x] Paste the output of the `Linter Eslint: Debug` command from the Command Palette below

```
Atom version: 1.25.0
linter-eslint version: 8.4.1
ESLint version: 4.19.1
Hours since last Atom restart: 0
Platform: darwin
Using local project ESLint from: /Users/bennyp/Documents/mws-restaurant-stage-1/node_modules/eslint
Current file's scopes: [
  "javascript",
  "program",
  "lexical_declaration",
  "variable_declarator",
  "arrow_function",
  "arrow_function",
  "parenthesized_expression",
  "sequence_expression"
]
linter-eslint configuration: {
  "disableWhenNoEslintConfig": false,
  "eslintrcPath": "/Users/bennyp/.config/eslintrc.json",
  "fixOnSave": true,
  "globalNodePath": "/Users/bennyp/.config/yarn/global",
  "ignoreFixableRulesWhileTyping": true,
  "lintHtmlFiles": true,
  "rulesToDisableWhileFixing": [
    "no-debugger"
  ],
  "scopes": [
    "source.js",
    "source.jsx",
    "source.js.jsx",
    "source.babel",
    "source.js-semantic",
    "text.html.basic"
  ],
  "useGlobalEslint": false,
  "showRuleIdInMessage": true,
  "advancedLocalNodeModules": "",
  "eslintRulesDirs": [],
  "disableEslintIgnore": false,
  "disableFSCache": false,
  "rulesToSilenceWhileTyping": []
}

```

You're welcome to try it in this repo: https://github.com/bennypowers/mws-restaurant-stage-1.git

skylize:
[...]

PR #1188, opened 2018-10-25
chore(readme): add info on linting other file types

trusktr:
I was scratching my head playing with `eslintrc` trying to figure why it wouldn't work, then I realized I had to add it to the scopes list when I stumbled on https://github.com/AtomLinter/linter-eslint/issues/1122

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,10 @@ installed
 for you. If you are using an alternative `linter-*` consumer feel free
 to disable the `linter` package.
 
-If you wish to lint files in JavaScript-derivative languages (like Typescript, Flow) with ESLint, you must add the scope name for that grammar to the `List of scopes to run ESLint on` option in `linter-eslint` Settings. For example, to lint TypeScript files, add `source.ts` to the list.
+If you wish to lint files in JavaScript-derivative languages (like Typescript,
+Flow) with ESLint, you must add the scope name for that grammar to the
+`List of scopes to run ESLint on` option in `linter-eslint` Settings. For
+example, to lint TypeScript files, add `source.ts` to the list.
 
 ## Use with plugins
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ installed
 for you. If you are using an alternative `linter-*` consumer feel free
 to disable the `linter` package.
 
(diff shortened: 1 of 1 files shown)


skylize:
Thanks for helping out! We can always benefit from better documentation. But this issue was fixed on Atom's side by changing Tree Sitter to use the old `source.js` scope name. 

Update to the newest [Release 1.32](http://blog.atom.io/2018/10/23/atom-1-32.html), which includes activating Tree Sitter by default and has no need for adding the `javascript` scope name that was required during the experimental stages.

skylize:
Sorry. I just actually read the commits, and realized I misunderstood what you were adding. 🤦‍♂️  because of the issue you linked to.

README.md
```diff
@@ -27,6 +27,8 @@ installed
 for you. If you are using an alternative `linter-*` consumer feel free
 to disable the `linter` package.
 
+> **Important!** If you wish to lint other files besides JavaScript files (f.e. `.js`), you need to add them to `List of scopes to run ESLint on` option in linter-eslint settings. For example, to lint TypeScript files, you'll need to add `source.ts` to the list.
```
skylize: Definitely do *not* think this qualifies for a bold "Important". This is intended for ESLint, which is in turn intended for JavaScript. Other languages are sort of an edge case. (Typescript has its own linter.)

skylize: This should go in A Note About Settings, not in Installation.

Arcanemagus:
```suggestion
> **Important!** If you wish to lint files besides JavaScript, you'll need to add them to the `List of scopes to run ESLint on` option in `linter-eslint`'s settings. For example, to lint TypeScript files, you'll need to add `source.ts` to the list.
```

skylize:
```suggestion
If you wish to lint files in JavaScript-derivative languages (like Typescript, Flow) with ESLint, you must add the scope name for that grammar to the `List of scopes to run ESLint on` option in `linter-eslint Settings`. For example, to lint TypeScript files, add `source.ts` to the list.
```

dakotahp:
@trusktr This PR is the only way I figured out how to get ESLint to run on TypeScript files. If you make these suggested changes then they can approve it for merging. Took me weeks to figure out how to do this and it's a shame that the readme doesn't already document this.

UziTech (approved):

PR #1194, opened 2018-11-05
Add source.flow to default scopes

IanVS:
Recent versions of atom do not add `source.js` in addition to `source.flow`,
treating flow as a completely different grammar from javascript, it seems.

This broke linting (along with other things) for flow users, unless
they added the flow source back to their config.  Instead, I think we should allow `linter-eslint` to "just work" for flow as well as regular javascript, which this
change will do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
         "source.js",
         "source.jsx",
         "source.js.jsx",
+        "source.flow",
         "source.babel",
         "source.js-semantic"
       ],


skylize:
I'm mixed in this. Flow seems just barely popular enough right now to validate this. But its future looks a bit grim. 

I certainly don't want people to just expect every transpiled variance

steelbrain:
Can we please get a release out with this included?

IanVS:
OK, finally got a release out.  Thanks for the prompt, @steelbrain.

PR #1212, opened 2018-12-30
v5.8.0

IanVS:
I no longer have permissions to push this straight to master, so trying a PR to see if we can get a release out.


IanVS:
@skylize, I think we might be able to get a release out if you approve this PR.

Issue #1195, opened 2018-11-12
Invalid position given by 'jsdoc/no-undefined-types'


thegeekajay:
ESLint returned a point that did not exist in the document being edited.
Rule: `jsdoc/no-undefined-types`
Requested start point: 715:0 - 715:0

Debug information:
```json
{
  "atomVersion": "1.32.0",
  "linterEslintVersion": "8.4.1",
  "linterEslintConfig": {
    "lintHtmlFiles": false,
    "useGlobalEslint": false,
    "showRuleIdInMessage": true,
    "disableWhenNoEslintConfig": true,
    "eslintrcPath": "",
    "globalNodePath": "",
    "advancedLocalNodeModules": "",
    "eslintRulesDirs": [],
    "disableEslintIgnore": false,
    "disableFSCache": false,
    "fixOnSave": false,
    "scopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic"
    ],
    "rulesToSilenceWhileTyping": [],
    "rulesToDisableWhileFixing": [],
    "ignoreFixableRulesWhileTyping": false
  },
  "eslintVersion": "5.9.0",
  "hoursSinceRestart": 18.6,
  "platform": "darwin",
  "eslintType": "local project",
  "eslintPath": "/Users/ganesha/Documents/ohcm/ohcm-fan-service/node_modules/eslint",
  "editorScopes": [
    "source.js",
    "keyword.control"
  ]
}
```

Issue #1192, opened 2018-10-31
Invalid position given by 'jsdoc/check-types'


brettz9:
ESLint returned a point that did not exist in the document being edited.
Rule: `jsdoc/check-types`
Requested start point: 10:0 - 10:0

Debug information:
```json
{
  "atomVersion": "1.32.0",
  "linterEslintVersion": "8.4.1",
  "linterEslintConfig": {
    "lintHtmlFiles": false,
    "useGlobalEslint": false,
    "showRuleIdInMessage": true,
    "disableWhenNoEslintConfig": true,
    "eslintrcPath": "",
    "globalNodePath": "",
    "advancedLocalNodeModules": "",
    "eslintRulesDirs": [],
    "disableEslintIgnore": false,
    "disableFSCache": false,
    "fixOnSave": false,
    "scopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic"
    ],
    "rulesToSilenceWhileTyping": [],
    "rulesToDisableWhileFixing": [],
    "ignoreFixableRulesWhileTyping": false
  },
  "eslintVersion": "5.8.0",
  "hoursSinceRestart": 3,
  "platform": "darwin",
  "eslintType": "local project",
  "eslintPath": "/Users/brett/svgedit/node_modules/eslint",
  "editorScopes": [
    "source.js",
    "comment.block"
  ]
}
```

FWIW, the sample I'm experimenting with is:

```js
/**
 * Namespaces or tools therefor
 * @module namespaces
 * @license MIT
*/

/**
 * @interface
 * @param {string} name The color name
 * @returns {object} A color object
 */
function Color (name) {}

/**
 * Function representing a transparent color.
 * @implements {Color}
 */
function TransparentColor (a) {}

export {Color, TransparentColor};
```

Should this instead be reported to eslint-plugin-jsdoc?

PR #1217, opened 2019-01-07
Fix behavior when endColumn or endLine is null

### LINKED ISSUE #1197: Invalid position given by 'jsdoc/no-undefined-types'
ESLint returned a point that did not exist in the document being edited.
Rule: `jsdoc/no-undefined-types`
Requested start point: 715:0 - 715:0

Debug information:
```json
{
  "atomVersion": "1.32.0",
  "linterEslintVersion": "8.4.1",
  "linterEslintConfig": {
    "lintHtmlFiles": false,
    "useGlobalEslint": false,
    "showRuleIdInMessage": true,
    "eslintrcPath": "",
    "globalNodePath": "",
    "eslintRulesDirs": [],
    "disableEslintIgnore": false,
    "disableFSCache": false,
    "fixOnSave": false,
    "scopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic"
    ],
    "rulesToSilenceWhileTyping": [],
    "rulesToDisableWhileFixing": [],
    "ignoreFixableRulesWhileTyping": false
  },
  "eslintVersion": "5.9.0",
  "hoursSinceRestart": 18.6,
  "platform": "darwin",
  "eslintType": "local project",
  "eslintPath": "/Users/ganesha/Documents/ohcm/ohcm-fan-service/node_modules/eslint",
  "editorScopes": [
    "source.js",
    "keyword.control"
  ]
}
```

mattlyons0:
Fixes #1197, #1196, #1195, #1192 

The JSDoc plugin returns `null` when there isn't a end column. The ESLint API docs specifically say the type of `endColumn` and `endLine` are optional or must be a `number`, therefore it is probably more correct and safer to do that check instead of a undefined check: https://eslint.org/docs/developer-guide/working-with-plugins

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -274,7 +274,7 @@ export async function processESLintMessages(messages, textEditor, showRule) {
      keep doing so in later uses.
      */
     const msgLine = line - 1
-    if (typeof endColumn !== 'undefined' && typeof endLine !== 'undefined') {
+    if (typeof endColumn === 'number' && typeof endLine === 'number') {
       eslintFullRange = true
       // Here we always want the column to be a number
       msgCol = Math.max(0, column - 1)
@@ -283,7 +283,7 @@ export async function processESLintMessages(messages, textEditor, showRule) {
     } else {
       // We want msgCol to remain undefined if it was initially so
       // `generateRange` will give us a range over the entire line
-      msgCol = typeof column !== 'undefined' ? column - 1 : column
+      msgCol = typeof column === 'number' ? column - 1 : column
     }
 
     let ret = {
--- a/spec/linter-eslint-spec.js
+++ b/spec/linter-eslint-spec.js
@@ -8,6 +8,8 @@ import rimraf from 'rimraf'
 import { beforeEach, it, fit } from 'jasmine-fix'
 import linterEslint from '../src/main'
 
+import { processESLintMessages } from '../src/helpers'
+
 const fixturesDir = path.join(__dirname, 'fixtures')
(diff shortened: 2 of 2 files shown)


src/helpers.js
```diff
@@ -274,7 +274,7 @@ export async function processESLintMessages(messages, textEditor, showRule) {
      keep doing so in later uses.
      */
     const msgLine = line - 1
-    if (typeof endColumn !== 'undefined' && typeof endLine !== 'undefined') {
+    if (typeof endColumn === 'number' && typeof endLine !== 'number') {
```
IanVS:
```suggestion
    if (typeof endColumn === 'number' && typeof endLine === 'number') {
```

mattlyons0: Woops, nice catch.
I've force pushed so a new commit isn't created just for that fix.

mattlyons0:
Nice catch. I was wondering why the tests were failing on the version I was running in terminal but not the one in atom, must have missed that char when copying between. Should pass now 🤞

IanVS:
Actually, one more thing before I merge this, do you think you could add a regression test, based on the results that the jsdoc plugin is giving?  Then we'll be sure not to break this in the future.

IanVS:
@mattlyons0 I haven't forgotten about this, just been busy.  I'll try to review in the next few days.  Thanks again.

mattlyons0:
Looks like you are following a different commit format now, updated.

PR #1245, opened 2019-02-26
Upgrade to Babel 7

Arcanemagus:
Now that we have a Babel 7 based transpiler available for Atom migrate our transpiler configuration over to that. This also includes moving from the deprecated and no longer maintained `babel-preset-env` to the current `@babel/preset-env`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,217 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Arcanemagus:
NOTE: I ran into some oddities when testing something unrelated where on this branch when running specs within Atom itself it would complain of missing modules. I haven't had a chance to look into whether this is a misconfiguration here or not.

PR #1305, opened 2019-09-23
fix(): infinite recursion when package.json in root and no eslint settings found

### LINKED ISSUE #1283: RangeError: Maximum call stack size exceeded
## Issue Type
Bug

## Issue Description
Infinite recursion for getConfigPath() in worker-helpers.js if package.json is present in root.
atom-linter.findCached will return package.json and getConfigPath will be called infinitely with root path (as it cannot go up anymore)

bogdandobanda:
Fixes #1283

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -6,6 +6,7 @@ import ChildProcess from 'child_process'
 import resolveEnv from 'resolve-env'
 import { findCached } from 'atom-linter'
 import getPath from 'consistent-path'
+import os from 'os'
 
 const Cache = {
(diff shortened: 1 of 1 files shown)


src/worker-helpers.js
```diff
@@ -20,6 +20,7 @@ const Cache = {
  * @return {string}      The cleaned path
  */
 const cleanPath = path => (path ? resolveEnv(fs.normalize(path)) : '')
+const rootPath = (process.platform === 'win32') ? process.cwd().split(Path.sep)[0] : '/'
```
aminya, line 23: Why don't you just use `process.cwd`?

PR #1358, opened 2020-07-14
add activation hook

aminya:
This speeds up loading of Atom, by deferring the loading of linter-eslint until the core Atom shell is loaded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -197,6 +197,9 @@
       }
     }
   },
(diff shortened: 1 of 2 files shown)


aminya:
@Arcanemagus could you approve this? 

I noticed that I do not have write-access to all repositories. Could you grant me access so I can maintain all the packages?

PR #1359, opened 2020-07-15
fix: expected a `length` to be non-negative error

aminya:
This fixes an error caused by Crypto API change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
     "atom-package-deps": "5.1.0",
     "babel-preset-env": "1.7.0",
     "consistent-path": "2.0.3",
-    "crypto-random-string": "2.0.0",
+    "crypto-random-string": "3.3.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -168,7 +168,7 @@
     "atom-package-deps": "5.1.0",
     "babel-preset-env": "1.7.0",
     "consistent-path": "2.0.3",
-    "crypto-random-string": "2.0.0",
+    "crypto-random-string": "^3.2.0",
```
Arcanemagus, line 171:
```suggestion
    "crypto-random-string": "3.2.0",
```
renovate will give us PRs when there are new versions, otherwise with the lockfile we would be stuck with the old version from when this PR was submitted until the range broke.

Arcanemagus (changes requested):
One minor change, other than that thanks for updating this!

PR #1360, opened 2020-07-15
Update "eslint" dependency

aminya:
This updates the Eslint dependency to 7

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
       "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
-      "dev": true,
       "requires": {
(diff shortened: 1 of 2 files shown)


aminya:
This is more complicated than I thought. CLIEngine is deprecated (still functional) in Eslint 7. If we want to upgrade the bundled Eslint we should use Linter and Eslint classes

https://eslint.org/docs/developer-guide/nodejs-api

src/worker-helpers.js
```diff
@@ -117,9 +117,12 @@ export function findESLintDirectory(modulesDir, config, projectPath, fallbackFor
  */
 export function getESLintFromDirectory(modulesDir, config, projectPath) {
   const { path: ESLintDirectory } = findESLintDirectory(modulesDir, config, projectPath)
+  console.log(ESLintDirectory)
   try {
     // eslint-disable-next-line import/no-dynamic-require
-    return require(ESLintDirectory)
+    const x = require(ESLintDirectory)
+    console.log(x)
```
aminya, line 124: This `require` doesn't work in the tests! 😶

aminya: @UziTech Does Jasmine change anything regarding `require` cache or something?

UziTech: Jasmine doesn't but atom-jasmine3-test-runner might

PR #1370, opened 2020-10-27
fix: fallback to other methods when global Eslint is not found

aminya:
This allows using Linter-Eslint even when the global Eslint is not found. Linting is better than no linting. 

Instead of `console.warn` we can use another method for informing the user.

### DIFF
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -47,10 +47,10 @@ function isDirectory(dirPath) {
   return isDir
 }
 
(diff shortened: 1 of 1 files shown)


Arcanemagus:
Using the global ESLint install is _heavily_ discouraged, and is disabled by default. If somebody has manually enabled it and hasn't set it up correctly they should have giant red flags warning them what a bad idea that is, not silently falling back to a working installation.

Thanks for taking the time to submit this though!

PR #1419, opened 2021-03-18
fix: fallback to getConfigPath

### LINKED ISSUE #1418: .conf/eslint/eslintrc.js ignored after update
## Issue Description (before update)
A project which has no .eslintrc file in its root folder used to eslint perfectly with having set:

- Global ESLint
  - [ ] Use global ESLint installation (unchecked)
  - [ ] Global Node Installation Path (empty)
  - [.conf/eslint/eslintrc.js] .eslint Path (set)

## Issue Description (after update)
Just did an update to 8.6.2. and got the following after restart:
"eslint parsing error: the keyword 'const' is reserved"

WTF? Obviously the file .conf/eslint/eslintrc.js is not considered any more.
Bug or Feature?

If Feature, then how do I make linter-eslint use .conf/eslint/eslintrc.js?
Adding .conf/eslint/ to "Uncommon/ESLint Rules Directories" does not change a thing
And no, I don't want another ".something" file polluting the projects root directory. :-)

Thanx 4 replies.

aminya:
Fixes #1418

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,8 +77,8 @@
           "order": 1
         },
         "eslintrcPath": {
-          "title": ".eslintrc Absolute Path",
-          "description": "The absolute path to eslint config file that is used when there's no config file in project",
+          "title": ".eslintrc Path",
+          "description": "It will only be used when there's no config file in project",
           "type": "string",
           "default": "",
           "order": 4
--- a/package.json
+++ b/package.json
@@ -77,8 +77,8 @@
           "order": 1
         },
         "eslintrcPath": {
-          "title": ".eslintrc Path",
-          "description": "It will only be used when there's no config file in project",
+          "title": ".eslintrc Absolute Path",
+          "description": "The absolute path to eslint config file that is used when there's no config file in project",
           "type": "string",
           "default": "",
           "order": 4
--- a/src/worker-helpers.js
+++ b/src/worker-helpers.js
@@ -139,7 +139,7 @@ export function getConfigForFile(eslint, filePath) {
 
 function getConfigPath(fileDir) {
   const configFile = findCached(fileDir, [
-    '.eslintrc.js', '.eslintrc.yaml', '.eslintrc.yml', '.eslintrc.json', '.eslintrc', 'package.json'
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -77,8 +77,8 @@
           "order": 1
         },
         "eslintrcPath": {
-          "title": ".eslintrc Path",
-          "description": "It will only be used when there's no config file in project",
+          "title": ".eslintrc Absolute Path",
+          "description": "The absolute path to eslint config file that is used when there's no config file in project",
```
UziTech, line 81: This doesn't currently need to be an absolute path. A relative path works for me.

aminya: Revertedd

src/worker-helpers.js
```diff
@@ -131,15 +131,15 @@ export function getConfigForFile(eslint, filePath) {
   try {
     return cli.getConfigForFile(filePath)
   } catch (e) {
-    // No configuration was found
-    return null
+    console.error('No configuration was found by eslint.CLIEngine, falling back to linter-eslint method.', e)
+    // eslint-disable-next-line no-use-before-define
```
UziTech, line 135: I feel like we should just move the function above this or disable the rule of we aren't going to follow it.

aminya: Yes, I was planning to use eslint-config-atomic. These strict rules set by Airbnb limit me without adding any benefits to the project.

But it seems that our tests rely on the eslint config in `package.json`. So, we need to fix that before changing the eslint config.

For now, I think disabling these rules doesn't hurt. Later we can get rid of them.

UziTech:
I don't think this is actually necessary. It seems like the old way was a subset of places checked by eslint's way of checking for config options. So if it doesn't find it with eslint's way it will never find it with the old way.

aminya:
If that's the case it is fine. But I want to cherry-pick the commit that requires an absolute path. Using a relative path means the user should put their config in that path for all the projects.

UziTech:
> If that's the case it is fine. But I want to cherry-pick the commit that requires an absolute path. Using a relative path means the user should put their config in that path for all the projects.

What if they want a different config file for each project but want it relative to the project folder.

aminya:
I am going to mention the need for using the config in all the projects.

PR #1422, opened 2021-03-24
Improve the tests

aminya:
- make tests independent of the project Eslint config:
	- Allows us to upgrade the Eslint config without being worried about the tests
	- Allows us to use eslint-config-atomic without breaking the tests

- use jasmine 3
- other fixes


aminya (commented):
Not sure why, but the tests never start on my local machine

UziTech:
> Not sure why, but the tests never start on my local machine

Atom caches the `atomTestRunner` in package.json so it is most likely not picking up the change in `atomTestRunner`. You have to close all instances of Atom then reopen to remove the cache.

aminya:
> > Not sure why, but the tests never start on my local machine
> 
> Atom caches the `atomTestRunner` in package.json so it is most likely not picking up the change in `atomTestRunner`. You have to close all instances of Atom then reopen to remove the cache.

The issue is `testPackages`. If I comment that out, it works fine.

aminya:
The failing tests are because of the way makeFixes is written
https://github.com/AtomLinter/linter-eslint/blob/1dca397d18795df5696eb9c4d6ac3a210a1b3ef9/spec/linter-eslint-spec.js#L88

spec/linter-eslint-spec.js
```diff
@@ -100,13 +109,14 @@ async function makeFixes(textEditor) {
...
   // Send off a fix request.
-  atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
+  await atom.commands.dispatch(atom.views.getView(textEditor), 'linter-eslint:fix-file')
 
   const notification = await notificationPromise
   expect(notification.getMessage()).toBe(expectedMessage)
   expect(notification.getType()).toBe('success')
 
   // After editor reloads, it should be safe for consuming test to resume.
+  buffer.reload()
```
aminya, line 119: Not sure if this is the right thing to do, but it fixes the tests

PR #1427, opened 2021-04-13
fix ignoreFixableRulesWhileTyping

jacekkopecky:
The setting `ignoreFixableRulesWhileTyping` currently doesn't work.

A `cliEngine` does not have `getRules` among its own properties, instead the function is a class method. Because of this, the `getRules` function in `worker-helpers.js` would just return an empty map and `ignoreFixableRulesWhileTyping` cannot know what rules to ignore.

This PR fixes that, now fixable rules are nicely ignored while typing.


jacekkopecky:
I've fixed “CI / CommitLint (pull_request)”; the problem with “CI / Lint (pull_request)” does not seem like something I can affect, please advise.

UziTech (approved):

UziTech:
The CI / Lint issue is a known issue unrelated to these changes.

aminya (approved):
Thanks for the contribution

